PR #6, opened 2023-05-05
PVNet2

dfulu:
# Pull Request
Updates for work done under PVNet2

## Description

- Updates to use pvnet datapipe from `ocf_datapipes` 
- Write a datamodule which uses `DataLoader2` and multiprocessing. Can either construct batches on the fly or load from premade batches
- Some model architecture upgrades (mainly around activation function, batchnorm, skip connections)
- Refactor the model so it is is composable and different model components can be plugged in and out
    - Include `encoders` module for model components which take 3D satellite/NWP data and encode it down to 1D vector
    - Include `linear_networks` module for model components which accept multiple 1D feature vectors from different sources and fuse them together into a prediction
- Add variations on the overall model skeleton
    -  `multimodal` - Original model skeleton. Encodes satellite and NWP and fuses them together with GSP history, solar coordinates, and ID embedding.
    - `deep_supervision` - Adds additional prediction heads to make predictions of GSP future based only on satellite and only on NWP
    - `weather residual` - Separates data sources so that GSP history, ID and solar coords are used to make initial prediction, and satellite and NWP are used to refine this prediction.
    - `nwp_weighting` - Completely different architecture where the prediction is a learned weighted mean of the dwsrf channel
- Improved logging during model training

## Checklist:


jacobbieker (approved):
Looks good overall! Did mostly skim over it, but yeah, looks good, and if it trains correctly, thats great! One thing to maybe add in a followup PR would be the HuggingFace integration with the model, wouldn't add it to this one as its already quite large, and it should be a pretty small change as well.

configs/datamodule/configuration/template_configuration.yaml
```diff
@@ -0,0 +1,98 @@
+general:
```
peterdudfield, line 1: Do we need all 4 of these configurations? Could you add alittle readme saying what they are and why?

configs/experiment/conv3d_sat_nwp.yaml
```diff
@@ -8,7 +8,7 @@ defaults:
   - override /model: conv3d_sat_nwp.yaml
   - override /datamodule: ocf_datapipes.yaml
   - override /callbacks: default.yaml
-  - override /logger: neptune.yaml
+#  - override /logger: neptune.yaml
```
peterdudfield, line 11: your not using neptune? are you using wandb?

configs/model/conv3d_sat_nwp_v6.yaml
```diff
@@ -0,0 +1,56 @@
+_target_: pvnet.models.multimodal.multimodal.Model
```
peterdudfield, line 1: do we need all 6? Maybe we could trim these?

pvnet/models/multimodal/deep_supervision.py
```diff
@@ -0,0 +1,329 @@
...
+        loss = loss / len(y_hats)
+
+        losses["MAE/train/multi-mode"] = loss
+
+        self._training_accumulate_log(batch, batch_idx, losses, y_hats["all"])
+
+        return loss
+
+
+if __name__ == "__main__":
```
peterdudfield, line 274: do you need this main in here? Move to a test?

pvnet/models/multimodal/multimodal.py
```diff
@@ -0,0 +1,280 @@
...
+            sun = self.sun_fc1(sun)
+            modes["sun"] = sun
+
+        out = self.output_network(modes)
+
+        return out
+
+
+if __name__ == "__main__":
+    from torch.optim import SGD
```
peterdudfield, line 226: similar to above, move to test if needed?

pvnet/models/multimodal/readme.md
peterdudfield, line 1: add something to this?

requirements.txt
```diff
@@ -6,20 +6,20 @@ matplotlib
...
+torch>=2.0
+pytorch-lightning>=2.0
 pytest
 pytest-cov
 flake8
 jedi
 tables
 tilemapbase
-neptune-client[pytorch-lightning]
+wandb
```
peterdudfield, line 17: for production, we might want to move some of these to extra requirments, but lets see

tests/data/sample_batches/train/000000.pt
peterdudfield, line 1: what is `.pt`?

peterdudfield:
> Ive left a few comments, but in general looks really good

I would try to get this merged, and then work on production as a separate branch

peterdudfield:
are the tests running automatically? Can you fix the pre-commit.ci?

configs/experiment/conv3d_sat_nwp.yaml
dfulu: Yeh, I made the switch to wandb

peterdudfield:
Could you also add a readme.md for the latest experiment you have done, doesnt have to be much text, but I find it really useful to have it near the code,

see example [here](https://github.com/openclimatefix/uk-pv-national-xg/tree/main/experiments)

tests/data/sample_batches/train/000000.pt
dfulu: I'm not actually sure if `.pt` is a distinct format. This file (along with all the pre-batched stuff) was saved with [torch.save()](https://pytorch.org/docs/stable/generated/torch.save.html) and they set the `.pt` extension for that. I went long with that convention.

dfulu: But the file contains an `ocf_datapipes` NumpyBatch

jacobbieker: .pt is just the default PyTorch extension for when it saves anything, usually Tensors

PR #10, opened 2023-05-19
Issue/production

dfulu:
# Pull Request

## Description

Add inference script for production

## How Has This Been Tested?

Test for script included in tests

## Checklist:


tests/data/sample_data/non_hrv_zeros.zarr/.zattrs
```diff
@@ -0,0 +1 @@
+{}
```
peterdudfield, line 1: Could you make this data on the fly, rather than committing it? 
In the past we commit these test zarr files but they quickly got out of data, and became annoying to update

peterdudfield:
Looks really great few things to do perhaps the merge
- create test zarrs on the file, rather than committing them
- add experiment folder with 001_pvnet_2.0 and a readme.md say roughly what this model and some top level results. This will be useful to reference in the future
- add an issue to fix the pre-comiit (or fix the pre commit now)

tests/data/sample_data/non_hrv_zeros.zarr/.zattrs
dfulu: How about a halfway solution where the coordinates are saved out as a zarr but the data itself is created on the fly?

This means we don't have to have a big list of x and y coordinates, variable names, timestamps, steps in `conftest.py` as well as some attributes we need to append to describe the coordinates. I've just committed this change, maybe have a look? 

I think it makes the conftest cleaner than it would be otherwise, and the saved data is pretty light with a total of 232KB for the NWP and satellite combined

dfulu:
@peterdudfield 

> * add experiment folder with 001_pvnet_2.0 and a readme.md say roughly what this model and some top level results. This will be useful to reference in the future

In the top level README there is a link at the bottom to an openly-viewable google doc that I'm maintaining with all the experimental results, including a high level summary at the top. Does that cover this, or do you still think the new folder is needed?

> * add an issue to fix the pre-comiit (or fix the pre commit now)

Currently working on these

peterdudfield:
> @peterdudfield
> 
> > * add experiment folder with 001_pvnet_2.0 and a readme.md say roughly what this model and some top level results. This will be useful to reference in the future
> 
> In the top level README there is a link at the bottom to an openly-viewable google doc that I'm maintaining with all the experimental results, including a high level summary at the top. Does that cover this, or do you still think the new folder is needed?
> 
> > * add an issue to fix the pre-comiit (or fix the pre commit now)
> 
> Currently working on these

Yea thats good for the moment, The google doc is very useful and nicely formatted. Something I quite like about having it in the code, means it is version controlled, and then very easy to see if we upgrade it. Perhaps we can tackle that later on

PR #42, opened 2023-06-27
New model

dfulu:
# Pull Request

## Description

Update app to use model trained to predict output noralised by effective capacity. Trained with constant 60 minute delay in satellite data as well.

## Checklist:


configs/datamodule/configuration/app_configuration.yaml
```diff
@@ -36,7 +36,7 @@ input_data:
 
   satellite:
     satellite_zarr_path: latest.zarr.zip
-    history_minutes: 120
+    history_minutes: 90
```
peterdudfield, line 39: how come this?

peterdudfield: perhaps it only uses 30 mins of satellite data, +60 minute of delat

dfulu: Mainly because I trained it on the [-90, -60] minute interval. It meant I could modify some preprepared batches rather than recreate them from scratch. Also this might have been bug since the [training config](https://github.com/openclimatefix/PVNet/blob/main/configs/datamodule/configuration/gcp_configuration.yaml) had this setting

dfulu: If we wanted to use the [-120, -60] interval we would need to retrain. So that might be the next iteration after this?

pvnet/app.py
```diff
@@ -63,7 +63,7 @@
 
 # Huggingfacehub model repo and commit
 model_name = "openclimatefix/pvnet_v2"
-model_version = "4a0510b498c55fee00defb385eec418d5712124c"
+model_version = "3d01665c89501da25e3acaaee6b669a15748708c"
```
peterdudfield, line 66: do you have some summary results of this model compared to the odd model? 
Perhaps comparing forecast horizon 0,1,2,4, and 8? This helps build confidnece about this new model

dfulu: Not yet. In the training/val losses it is slightly harder to compare since the new model uses effective capacity as a normalising factor and the older one uses installed capacity as the normalising factor. So the losses aren't super easily comparable. By making a rough estimate that in the validation period, the average effective/installed capacity is about 0.95, you can estimate the effect of the different normalising factor. Looking at wandb logs and using that it seems like this new model is marginally better in the validation period of 2021. However, I'd expect the gap to be bigger now as the ratio grows with time. I could compile this table  properly early next week.

When I get the chance ,I could run them both on the production dataset. However, I'll need to gather a new production dataset because of the issue you mentioned about the satellite history not being long enough. So it'll take a week or two to have some data for this.

peterdudfield: Ok that makes sense. 

it's a lesson learnt that we shuold make sure we run our metrics on non normalized values as well. This means if we change are normalization then we can still comapre

PR #44, opened 2023-07-07
Add quantile regression support

dfulu:
# Pull Request

## Description

Add option to train PVNet with quantile regression. An example training run using these new changes is [hosted on wandb](https://wandb.ai/openclimatefix/pvnet2.1/runs/0vvcz7f3). Figure at bottom shows some results from the validation set.

These changes are:
- Add quantile loss as a loss function to minimise on, whilst still tracking MAE/MSE of the 0.5 quantile
- Add plotting functionality to plot all predicted quantiles if used
- Modify configs and training script to choose correct metric to monitor for early stopping and checkpointing
- Add tests for quantile regression option in model

## Checklist:

media_images_val_forecast_samples_batch_idx_1_573_8b7bed89957cfd499875


peterdudfield:
Amazing, well done for doing this!!!! 

Do you fancy upgradeing to nowcasint_datamodel=1.4.14. and then in your[convert](https://github.com/openclimatefix/PVNet/blob/main/pvnet/app.py#L91) anding a `properties' field. 
This field should be a dictionary like  
`forecast_value_sql.properties = {'plevel_10':forecast_value.plevel_10, 'plevel_90':forecast_value.plevel_90}

dfulu:
> Amazing, well done for doing this!!!!
> 
> Do you fancy upgradeing to nowcasint_datamodel=1.4.14. and then in your[convert](https://github.com/openclimatefix/PVNet/blob/main/pvnet/app.py#L91) anding a `properties' field. This field should be a dictionary like `forecast_value_sql.properties = {'plevel_10':forecast_value.plevel_10, 'plevel_90':forecast_value.plevel_90}

@peterdudfield I was updating these libraries in a separate issue #45.  I'll work on this after that pull request is merged to main and then merge it into this.

dfulu:
> Also have you added stuff to the PVnet public document? WOuld great to get a sense of how the MAE has changed? and how good the quantiles are?

@peterdudfield I haven't yet, but can do. At least the wandb run linked above shows the MAEs. In short the MAE for this model trained on quantile loss is the same as when the model is trained only to predict MAE. The get identical scores (allowing for some jitter in the training/validation curves).

dfulu:
@peterdudfield Also, this isn't quite a full solution yet since we don't have a way to sum the quantiles to get a probabilistic forecast for the national sum. 

For example, if could sum all of the 10th percentile lines from all GSPs, and all the 90th percentile lines, and use this for the range for the PVNet national estimate. However, we'd end up with to large of a 10-90th percentile range for the national sum. This option would essentially to assume that the prediction errors will all be perfectly correlated. 

Alternatively, we might be able to do something similar to adding the uncertainties in quadrature. But this assumes that the errors in the prediction are perfectly uncorrelated, and so we'd end up with a 10-90th percentile range which is too small.

The reality is likely to be somewhere in the middle of these two, which partial correlation of errors. I think the best option might be to train another very small model to estimate the quantiles of the national sum from the output of this model

peterdudfield:
> @peterdudfield Also, this isn't quite a full solution yet since we don't have a way to sum the quantiles to get a probabilistic forecast for the national sum.
> 
> For example, if could sum all of the 10th percentile lines from all GSPs, and all the 90th percentile lines, and use this for the range for the PVNet national estimate. However, we'd end up with to large of a 10-90th percentile range for the national sum. This option would essentially to assume that the prediction errors will all be perfectly correlated.
> 
> Alternatively, we might be able to do something similar to adding the uncertainties in quadrature. But this assumes that the errors in the prediction are perfectly uncorrelated, and so we'd end up with a 10-90th percentile range which is too small.
> 
> The reality is likely to be somewhere in the middle of these two, which partial correlation of errors. I think the best option might be to train another very small model to estimate the quantiles of the national sum from the output of this model

Yea, perhaps before training a model, could first do some analysis on how they are correlated?

peterdudfield:
@dantravers might actually know this. 

the problem is how to add up GSP uncertainites to make the national one. Assuming normal distributions
1. If uncorreclated, can square them and sum, then square root
2. if correctted, can just up them up.

one way to perhaps do it would be to do a mixture of 1. and 2. and see what is best. 

As you suggested @dfulu could do a seperatel model and train this, but perhaps a simple way first is a good start

dfulu:
@peterdudfield  it will take a bit of time to do the analysis. The biggest component of this is preparing the input dataset since it takes 317 samples at the same timestamp to give us 1 data point for the correlations or to train a model. I'll start working on this on a separate branch

dantravers:
On adding up the uncertainties: 
Well - as luck should have it, I was just talking about this!  I think the right thing to do is: 
Create a correlation matrix C, where each C_i_j is the correlation of the timeseries GSPerror_i, and GSPerror_j, where i, j are the index of the GSP (from 1 to n). C is nxn, where there are n GSPs.
GSPerror timeseries is formed by working out the forecast minus the actual for the GSP for all historical periods we have (for the horizon we are looking at).  We could create this based on some backetst / cross validation period.   (I've actually calculated this for many sites in GSPs in my PhD stuff, but there are some gaps in sites, so it would be better to do for all GSPs with the same model I think. )

Then the (median-quantile) (for each up and down quantile) would be Q@C@Qt, where Q is the nx1 vector of (median-quantile level) for each GSP, Qt is Q transpose, and @ is dot product.

dfulu:
To keep this up date with conversations in person. I think @dantravers's idea is a pretty good approximation. However, I think its would be almost the same amount of work to go to what feels like a "full" solution of training a model on all of the GSP predictions to estimate the quantiles. That is the current goal

peterdudfield:
Is it worth adding a test, to check the probablsitic works?

tests/models/multimodal/test_multimodal.py
```diff
@@ -24,3 +30,20 @@ def test_model_backward(multimodal_model, sample_batch):
 
     # Backwards on sum drives sum to zero
     y.sum().backward()
+
+
+def test_quantile_model_forward(multimodal_quantile_model, sample_batch):
```
dfulu, line 35: @peterdudfield there already are a couple of tests here

PR #45, opened 2023-07-07
Update app data input to allow 15-minute satellite fallback

dfulu:
# Pull Request

## Description

Update the app to allow 15-minute satellite data to be used in app when the 5-minute data is not available. See openclimatefix/ocf_datapipes#209 for related update to ocf_datapipes.

We don't store 15-minute data in the database unless the 5 minute data is not available. I tested this locally by downloading the 5-minute data. I modified this downloaded data so that the newest time stamp was 2 hours before present, and saved this as `latest.zarr.zip`. I also took the downloaded 5-minute data, filtered it to 15 minutely and named it `latest_15.zarr.zip`. I then used these to run the `app.py` and it ran successfully.

Changes:
- Update requirements
- Download 15-minute data if available

## Checklist:

### DIFF
--- a/pvnet/app.py
+++ b/pvnet/app.py
@@ -238,7 +238,7 @@ def app(
     logger.info("Downloading zipped satellite data")
     fs = fsspec.open(os.environ["SATELLITE_ZARR_PATH"]).fs
     fs.get(os.environ["SATELLITE_ZARR_PATH"], "latest.zarr.zip")
-    
+
     # Also download 15-minute satellite if it exists
     sat_latest_15 = os.environ["SATELLITE_ZARR_PATH"].replace(".zarr.zip", "_15.zarr.zip")
     if fs.exists(sat_latest_15):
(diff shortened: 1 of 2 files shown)


pvnet/app.py
```diff
@@ -239,6 +239,12 @@ def app(
     fs = fsspec.open(os.environ["SATELLITE_ZARR_PATH"]).fs
     fs.get(os.environ["SATELLITE_ZARR_PATH"], "latest.zarr.zip")
 
+    # Also download 15-minute satellite if it exists
```
peterdudfield, line 242: ah, I think you dont need to do this, because we have covered this in `ocf_datapipes` 
What do you think?

peterdudfield: Ah I think I miss read this, 

This is a good solution! One quick question, what happens if that 15 mintue file isnt there?

peterdudfield (approved):

Issue #53, opened 2023-07-17
Test using GSP PVLive inputs


dfulu:
Currently the model does not use any PV measurement as inputs. This is because we only have the updated PVlive estimates for each GSP and not the initial intra-day estimates. 

The updated estimates are likely to be different from the intraday estimates, but we are not sure by how much. To test their usefulness we should:

- Train a model with updated PVLive GSP inputs
- Test the trained model in the production environment with intra-day PVLive GSP inputs

dfulu:
The validation results for the current model (no hist) and the new model which also used uses PVLive GSP live inputs is shown here. Note that this validation data also used the updated PVLive data as inputs so the performance of the new model may differ in production

![W B Chart 17_07_2023, 10_00_48](https://github.com/openclimatefix/PVNet/assets/41546094/a9c7baab-8c97-4463-b62f-817013136edc)

peterdudfield:
Nice this is really good

dfulu:
<img width="768" alt="Screenshot 2023-08-08 at 15 54 42" src="https://github.com/openclimatefix/PVNet/assets/41546094/22272e28-6720-4224-8795-50638d1cd77a">

Above is the results for this experiment evaluated over the period 18 July - 5th August 2023 on the national sum of GSP predictions. Using intraday PVLive inputs to the model trained on updated PVLive inputs increased the accuracy of the short term predictions but decreased the accuracy after about an hour.

peterdudfield:
Ill just keep this open, so we remember the different between dev and production for the moment. 

One other way we could carry on here, is to deploy model PVlive for the first 1 hour, then switch to with out. Its a bit messy though

dfulu:
Yeh, that could work, but I'd prefer to keep it all in one model ideally. I think we should wait to see the results of #68. Option 2 of that should be reasonably quick to implement

dfulu:
Closing this, as it has become apparent that using PVLive inputs is not going to be possible and neither dev or production use it anymore

PR #54, opened 2023-07-17
Issue/intraday input model

dfulu:
# Pull Request

## Description

- Update model to use PVLive intraday inputs. This will help to test #53
- Also nicer formatting of model config uploaded to huggingface
- Add more docs to model upload script

## Checklist:

### DIFF
--- a/pvnet/app.py
+++ b/pvnet/app.py
@@ -62,7 +62,7 @@ batch_size = 10
 
 # Huggingfacehub model repo and commit
 model_name = "openclimatefix/pvnet_v2"
-model_version = "3d01665c89501da25e3acaaee6b669a15748708c"
+model_version = "898630f3f8cd4e8506525d813dd61c6d8de86144"
 
 model_name_ocf_db = "pvnet_v2"
 use_adjuster = os.getenv("USE_ADJUSTER", "True").lower() == "true"
--- a/pvnet/models/base_model.py
+++ b/pvnet/models/base_model.py
@@ -126,7 +126,7 @@ class PVNetModelHubMixin(PyTorchModelHubMixin):
 
         # saving config
         if isinstance(config, dict):
-            (save_directory / CONFIG_NAME).write_text(json.dumps(config))
+            (save_directory / CONFIG_NAME).write_text(json.dumps(config, indent=4))
 
         # Creating and saving model card.
         card_data = ModelCardData(language="en", license="mit", library_name="pytorch")
--- a/scripts/checkpoint_to_huggingface.py
+++ b/scripts/checkpoint_to_huggingface.py
@@ -1,4 +1,10 @@
-"""Command line tool to push locally save model checkpoints to huggingface"""
+"""Command line tool to push locally save model checkpoints to huggingface
+
+use:
+python checkpoint_to_huggingface.py "path/to/model/checkpoints" \
+    --local-path="~/tmp/this_model" \
+    --no-push-to-hub
+"""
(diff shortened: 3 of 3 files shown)


peterdudfield (changes requested):
Should there be something where the production config has changed to the inference knows to load in GSP?

Would be great to add something to the google doc. Personally, (and I know ive had push back before), I would prefer a experiments folder and add a small markdown with the results in it, and a small explation of what was done. 

Perhaps the thing to do its both
- General google doc that has lots of experiments in it
- markdown for each experiment example what has happened

What do you think

dfulu:
In production, the GSP data is already loaded by datapipes. Previously the model was just ignoring it in the batch. It's so little data that I didn't include the option not to load it, and we need to load some of the GSP data anyway to get the locations and the capacities. 

Personally, I prefer the google doc. I like to have a lot of graphs which would be extra data we have to include in the library and would make it bigger than necessary. I also like including ad-hoc analyses of the inputs and like having that alongside the experiments since it informs them, and the same ad-hoc analysis can inform multiple experiments. I've also found that experiments quickly go out of date; for example, everything that was done before the location bug fix isn't informative anymore. The production results before the zig-zag issue was solved, and the national results from before we solved the issue of the GSPs being shuffled by mutli-processing are also out of date. I think all of those are easier to keep up to date and to remove the invalidated experiments inside the google doc.

I'll add this new experiment to the google doc, but I'd like to wait until we have results from production to do the proper write-up of it. There is actually an [older experiment there](https://docs.google.com/document/d/1fbkfkBzp16WbnCg7RDuRDvgzInA6XQu3xh4NCjV-WDA#heading=h.wy7vb046vr23) which compared including the GSP or not. I'll probably just update that when the new results are in.

peterdudfield:
> In production, the GSP data is already loaded by datapipes. Previously the model was just ignoring it in the batch. It's so little data that I didn't include the option not to load it, and we need to load some of the GSP data anyway to get the locations and the capacities.
> 
> Personally, I prefer the google doc. I like to have a lot of graphs which would be extra data we have to include in the library and would make it bigger than necessary. I also like including ad-hoc analyses of the inputs and like having that alongside the experiments since it informs them, and the same ad-hoc analysis can inform multiple experiments. I've also found that experiments quickly go out of date; for example, everything that was done before the location bug fix isn't informative anymore. The production results before the zig-zag issue was solved, and the national results from before we solved the issue of the GSPs being shuffled by mutli-processing are also out of date. I think all of those are easier to keep up to date and to remove the invalidated experiments inside the google doc.
> 
> I'll add this new experiment to the google doc, but I'd like to wait until we have results from production to do the proper write-up of it. There is actually an [older experiment there](https://docs.google.com/document/d/1fbkfkBzp16WbnCg7RDuRDvgzInA6XQu3xh4NCjV-WDA#heading=h.wy7vb046vr23) which compared including the GSP or not. I'll probably just update that when the new results are in.

Ok, stick to google doc, and as you say, start a section for this bit. 

OK so `ocf_datapipes` uses it anyway, but shouldnt there be a bit of code in the model that turns [this](https://github.com/openclimatefix/PVNet/blob/main/pvnet/models/multimodal/multimodal.py#L50) on (and it was off before)

dfulu:
Yes there is, but it is in the model config which is saved alongside the model weights on huggingface. I thought that was nicer to link them up there and have no doubt as to which config is required for each checkpoint

peterdudfield (approved):

PR #69, opened 2023-08-11
make saving more reusable

dfulu:
# Pull Request

## Description

Very minor change to add additional parameter so the save function can be reused in pvnet_summation


jacobbieker (approved):
LGTM! Thanks!

PR #71, opened 2023-08-14
Use summation model in app

dfulu:
# Pull Request

## Description

- Add the summation model to the app
- Update tests

## Checklist:


pvnet/app.py
```diff
@@ -234,6 +248,9 @@ def app(
         .effective_capacity_mwp
     )
 
+    # National capacity is needed if using summation model
+    national_capacity = gsp_capacities.sum().item()
```
peterdudfield, line 252: should we take the sum here or should we take the national one from the database?

pvnet/app.py
```diff
@@ -284,15 +301,36 @@ def app(
...
+        ):
+            warnings.warn(
+                f"The PVNet version running in this app is "
+                f"{model_name}/{model_version}."
+                f"The summation model running in this app was trained on outputs from PVNet "
+                f"version {summation_model.model_name}/{summation_model.model_version}. "
+                f"Combining these models may lead to an error if the shape of PVNet output doesn't "
+                f"match the expected shape of the summation model. Combining may lead to "
+                f"unreliable results even if the shapes match."
+            )
```
peterdudfield, line 327: nice

pvnet/app.py
```diff
@@ -313,19 +351,27 @@ def app(
             elevation = elevation[:, -preds.shape[1] :]
             sun_down_mask = elevation < MIN_DAY_ELEVATION
 
-            # Zero out after sundown
-            preds[sun_down_mask] = 0
```
peterdudfield, line 317: why remove this?

peterdudfield: ah i see now, for the below to do the sun masking. 
Q: do we need to sun mask on national if we have done it on gsp?

pvnet/app.py
```diff
@@ -346,35 +392,95 @@ def app(
...
+        # Multiply normalised forecasts by capacities and clip negatives
+        da_abs_national = da_normed_national.clip(0, None) * national_capacity
+
+        # Apply sundown mask - All GSPs must be masked to mask national
+        da_abs_national = da_abs_national.where(~da_sundown_mask.all(dim="gsp_id")).fillna(0.0)
+
+        da_abs_all = xr.concat([da_abs_national, da_abs], dim="gsp_id")
+
+    else:
+        logger.info("Using simple sum")
```
peterdudfield, line 471: could you expand comment, to make it more clear

peterdudfield (approved):
Looks good!!! well done. Does it work when you run it locally?

dfulu:
Yeh it runs locally, here's this mornings forecast for comparison
Screenshot 2023-08-15 at 09 53 22

pvnet/app.py
dfulu: I had thought about that, but actually when I ran the line `ds_gsp = next(iter(OpenGSPFromDatabase()))`, `ds_gsp` did not have any information for national - no`gsp_id=0`

pvnet/app.py
dfulu: Yeh I think we do. The summation model will occassionally predict positive values during the night, mainly for the 90th percentile line. So it does need to be masked. Also the summation model was trained on non-sun-masked GSP inputs. To keep things the same I don't mask the GSP inputs fed into the summation model here either

pvnet/app.py
dfulu: I've added a commit to pull the national capacity from the database

pvnet/app.py
```diff
@@ -398,4 +506,4 @@ def app(
 
 
 if __name__ == "__main__":
-    typer.run(app)
+    2  # typer.run(app)
```
peterdudfield, line 509: This means it doesnt run anymore @dfulu , can you put this back in?>

PR #74, opened 2023-08-16
add default models to docker image

dfulu:
# Pull Request

## Description

Add default models to the docker image. The models are on the order of 100MB in size for pvnet and the summation model. The models will be saved to the on-disk model cache and this means we do not need to download them from huggingface each time we run the app.

This seems neater and makes us less susceptible to rare huggingface outages, as happened today.

## Checklist:


peterdudfield (approved):

peterdudfield:
Did you try it locally? did it work?

Issue #76, opened 2023-08-16
Upgrade dm


peterdudfield:
Could we upgrade datamodel to 1.5.10?

I tried to do it quickly, but then pvnet summationan was tagged to a specific version.

peterdudfield:
@dfulu would you be able to have a go at this?

PR #78, opened 2023-08-17
Update requirements

dfulu:
# Pull Request

## Description

Update requirements. Solves #76


peterdudfield (approved):
Even thought tests didnt run, we hope with this new release it should fix. 

Problem here is pvnet and pvnet summation depend on each other. But we can address this later

PR #81, opened 2023-08-17
model card uploaded to huggingface with model

dfulu:
# Pull Request

## Description

Fix bug which was not updating the model card on huggingface when a new model is uploaded. Now model and card are updated simultaneously and makes it easier to link the model to a specific run on wandb

Fixes #

## How Has This Been Tested?

Tested locally from command line. These functions are only ever intended to be run from command line


peterdudfield (approved):
Ill have to take your word on this, that it works

PR #88, opened 2023-10-24
Add site-level PV input subnetwork

dfulu:
# Pull Request

Add support to the multimodal model to optionally include an encoder for system-level PV inputs

- Add a couple of simple PV site-level encoder architectures
- Add PV site encoder to multimodal model
- Slight refactor of multimodal model and configs
- Update tests

## Checklist:


pvnet/models/multimodal/multimodal.py
```diff
@@ -56,105 +52,79 @@ def __init__(
...
-        source_dropout=0.0,
+        pv_history_minutes: Optional[int] = None,
         optimizer: AbstractOptimizer = pvnet.optimizers.Adam(),
     ):
-        """Neural network which combines information from different sources.
-
-        Args:
-            output_quantiles: A list of float (0.0, 1.0) quantiles to predict values for. If set to
-                None the output is a single value.
-            image_encoder: Pytorch Module class used to encode the NWP data (and satellite data
```
jacobbieker, line 71: Why would this be removed? I think having the options described can still be quite helpful

pvnet/models/multimodal/multimodal.py
```diff
@@ -56,105 +52,79 @@ def __init__(
...
-                image_size_pixels=nwp_image_size_pixels,
-                in_channels=number_nwp_channels + add_image_embedding_channel,
-                out_features=encoder_out_features,
-                **encoder_kwargs,
+                in_channels=nwp_encoder.keywords["in_channels"] + add_image_embedding_channel,
             )
             if add_image_embedding_channel:
-                self.nwp_embed = ImageEmbedding(318, nwp_sequence_len, nwp_image_size_pixels)
+                self.nwp_embed = ImageEmbedding(
+                    318, nwp_sequence_len, self.nwp_encoder.image_size_pixels
```
jacobbieker, line 106: Why is the embedding 318? Just curious

jacobbieker (approved):
LGTM!

pvnet/models/multimodal/multimodal.py
dfulu: Oops thanks, I meant to go back and update this

pvnet/models/multimodal/multimodal.py
dfulu: The nn.Embedding layer used under the hood inside ImageEmbedding expects an ID in the range [0, 1, ...., n-1] when you specify that the number of embeddings is n. The GSP IDs are 1-317 so I set n=318. So the embedding for ID 0 is wasted. Alternatively I could be passing in (GSP ID - 1) instead of (GSP ID).

 I think the 318 solution was inherited from PVNet  v1, and I didn't know the above when I started updating to v2. It would break our current production model to update it now. But I might do so in the future to clean it up

PR #105, opened 2023-12-13
Fixed error caused by incorrect call to common_metrics in validation step

confusedmatrix:
# Pull Request

## Description

This PR fixes a typo in a parameter name when calling `common_metrics` in the validation step during training of PVNet. Additionally, as `common_metrics` expects np.ndarrays as input, the input predictions and target values are cast from torch Tensors to np.ndarrays. 

This PR works in conjunction with https://github.com/openclimatefix/ocf-ml-metrics/pull/13 to provide correct logging of validation metrics during training using multiple forecast horizons.

## How Has This Been Tested?

Tested by training locally over pre-made data batches

- [x] Yes

_If your changes affect data processing, have you plotted any changes? i.e. have you done a quick sanity check?_

- [ ] Yes

## Checklist:


peterdudfield (approved):

peterdudfield:
@dfulu this looks good to go?

dfulu (changes requested):
If I'm correct I think that common_metrics returns scalar values for the rmse and mae. `mse_each_step` is expected  to be an array of mse values for each forecast step. Same with `mae_each_step`. So I think there is still some work to do to fix this

peterdudfield:
> If I'm correct I think that common_metrics returns scalar values for the rmse and mae. `mse_each_step` is expected to be an array of mse values for each forecast step. Same with `mae_each_step`. So I think there is still some work to do to fix this

but this is about the inputs? 
Also perhaps a test for this would be a good way to check this?

confusedmatrix:
> > If I'm correct I think that common_metrics returns scalar values for the rmse and mae. `mse_each_step` is expected to be an array of mse values for each forecast step. Same with `mae_each_step`. So I think there is still some work to do to fix this
> 
> but this is about the inputs? Also perhaps a test for this would be a good way to check this?

Yep - the PR in ocf-ml-metrics (here: https://github.com/openclimatefix/ocf-ml-metrics/pull/13) updates the common_metrics function to handle scalar and array return types + includes tests for both cases

dfulu:
Thanks @confusedmatrix, I merged your changes to ocf-ml-metrics, so once the build and pip release is done we should update the requirements here and that should be it.

Also @peterdudfield we probably should have a full end2end test for the PVNet training, loop, but I'd say lets do that in a separate issue

PR #117, opened 2024-01-08
Fix test batches and make weighted loss general to device

dfulu:
- Fix batches for datapipes 3.1.5
- Minor fix to make sure the weighted loss will work regardless of the device used


jacobbieker (approved):
LGTM!

PR #145, opened 2024-02-26
Add unimodal teacher method to train the multimodal model

dfulu:
# Pull Request

## Description

- Add in support for training the multimodal model via unimodal-teacher [1]
- Add tests for unimodal teacher model
- Add in support to convert the unimodal-teacher into a multimodal model and to push this to huggingface for production
- Also accidentally included another NWP / satellite encoder and tests
- Also included method to adapt batches which are too big either spatially or in time for easier experimentation
- Also light refactoring and removing duplicated code

[1] https://arxiv.org/pdf/2305.01233.pdf

## Checklist:


pvnet/models/multimodal/encoders/basic_blocks.py
```diff
@@ -78,6 +79,62 @@ def forward(self, x):
         return self.model(x) + x
 
 
+class ResidualConv3dBlock2(nn.Module):
```
jacobbieker, line 82: Could we use a different name?

pvnet/models/multimodal/encoders/encoders3d.py
```diff
@@ -201,6 +202,84 @@ def forward(self, x):
         return out
 
 
+class ResConv3DNet2(AbstractNWPSatelliteEncoder):
```
jacobbieker, line 205: Same here, could we use a different name? Is there some architectural thing that we could include in the name so it's not something2?

jacobbieker (approved):
LGTM! Great work! Only comments on naming the new architecture blocks, would be nice to differentiate them more from the original ones here.

pvnet/models/multimodal/encoders/encoders3d.py
dfulu: The reason for this was that the structure of this is the same as the `ResFCNet2` network. And `ResFCNet2` and `ResFCNet` differed by activation functions and dropout I think. So yeh we could combine them. I'd rather do that in a second pull request than tag it on to this though

PR #151, opened 2024-02-29
Remove old architectures

dfulu:
# Pull Request

## Description

Removes some old architectures which we no longer support and are no longer used. These have never been used in production, have not been trained in a year and are no longer functional as they weren't maintained through updates

For cleanliness mostly


peterdudfield (approved):
nice, great to clean that up. 

Just make sure the tests still pass

PR #163, opened 2024-03-19
Remove references to old datamodule config parameters

Sukh-P:
# Pull Request

## Description
Removing references in the example config files to old datamodule config parameters (`n_train_data` and `n_val_data`) which are no longer used in the [datamodule class](https://github.com/openclimatefix/PVNet/blob/main/pvnet/data/base.py#L8) (limiting number of batches is done through the trainer config parameters instead i.e `limit_train_batches`) 

This is to make it slightly clearer when looking at the examples what parameters are still in use. 

## Checklist:

### DIFF
--- a/configs.example/experiment/example_simple.yaml
+++ b/configs.example/experiment/example_simple.yaml
@@ -22,8 +22,6 @@ trainer:
   max_epochs: 2
 
 datamodule:
-  n_train_data: 2
-  n_val_data: 2
-  fake_data: 1
+  batch_size: 32
 
 validate_only: "1" # by putting this key in the config file, the model does not get trained.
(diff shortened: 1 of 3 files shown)


configs.example/experiment/example_simple.yaml
```diff
@@ -22,8 +22,6 @@ trainer:
   max_epochs: 2
 
 datamodule:
-  n_train_data: 2
-  n_val_data: 2
-  fake_data: 1
+  batch_size: 32
```
Sukh-P, line 25: Put an arbitrary example parameter and value here instead

jacobbieker (approved):

PR #164, opened 2024-03-19
Add some ignored paths to github workflow

Sukh-P:
# Pull Request

## Description
To prevent github workflows such as versioning updates/ releases triggering when making changes to non python code files some ignored paths have been added  

## How Has This Been Tested?
Will test this through the automated checks that run with each commit and will make a small change to some config examples to see if it works

## Checklist:

### DIFF
--- a/configs.example/experiment/example_simple.yaml
+++ b/configs.example/experiment/example_simple.yaml
@@ -22,6 +22,6 @@ trainer:
   max_epochs: 2
 
 datamodule:
-  batch_size: 32
+  batch_size: 16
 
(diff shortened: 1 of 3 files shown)


configs.example/experiment/example_simple.yaml
```diff
@@ -22,6 +22,6 @@ trainer:
   max_epochs: 2
 
 datamodule:
-  batch_size: 32
+  batch_size: 16
```
Sukh-P, line 25: This is more as a test to see if the ignored paths change above works

jacobbieker (approved):
LGTM! Thanks!

PR #175, opened 2024-04-12
Add first draft backtest script

dfulu:
## Description

Add first draft of a backtest script. This is for running PVNet and the summation model for the UK GSP and national level predictions. Also includes some minor tidying.

- [x] TODO test locally that it works

## How Has This Been Tested?

This has been tested locally. I ran the script with the user variables set to the values they are set to in the script. This was two init-times and using the PVNet model and summation model which are currently on the dev branch. After running the backtest for these two init-times I checked them against the PVLive values and they seemed reasonable

 To test with pytest would require a huge amount of build up and teardown to get it to work

## Checklist:


scripts/backtest_uk_gsp.py
```diff
@@ -0,0 +1,467 @@
...
+- This script uses hydra to construct the config, just like in `run.py`. So you need to make sure
+  that the data config is set up appropriate for the model being run in this script
+- The PVNet and summation model checkpoints; the time range over which to make predictions are made;
+  and the output directory where the results near the top of the script as hard coded user
+  variables. These should be changed.
+
+
+```
+python backtest.py
+```
```
peterdudfield, line 15: nice, thank you

peterdudfield:
Yea, it would be trickly to test with `pytest`, as you need a fair amount of fake data e.t.c. Its probably something to do in the future, but this is a great first step

scripts/backtest_uk_gsp.py
```diff
@@ -0,0 +1,467 @@
...
+from omegaconf import DictConfig
+
+# TODO: Having this script rely on pvnet_app sets up a circular dependency. The function
+# `preds_to_dataarray()` should probably be moved here
+from pvnet_app.utils import preds_to_dataarray
+from torch.utils.data import DataLoader
+from torch.utils.data.datapipes.iter import IterableWrapper
+from tqdm import tqdm
+
+from pvnet.load_model import get_model_from_checkpoints
```
Sukh-P, line 60: This may just be me missing something but where are the `load_model` file and the `get_model_from_checkpoints` function referenced in this import? Thought maybe these were meant to be included with this PR?

dfulu: Forgot to add. Done now

peterdudfield:
This ready to go @Sukh-P ?

Sukh-P (approved):
Yep I think this is good to go, I have tested it as well in part

PR #192, opened 2024-05-15
Add site backtest script

Sukh-P:
# Pull Request

## Description

Adding a site PVNet backtest script which has been modified from the [original script](https://github.com/openclimatefix/PVNet/commit/860e5e08d8955241b493f2a4c3095dce05fa7167) to be able to create a backtest for specific PV sites 

## How Has This Been Tested?

This has been tested by the running the script and verifying the outputs 
## Checklist:

### DIFF
--- a/pvnet/utils.py
+++ b/pvnet/utils.py
@@ -68,6 +68,31 @@ class GSPLocationLookup:
             id=gsp_id,
         )
 
+class SiteLocationLookup:
+
+    def __init__(self, long: xr.DataArray, lat: xr.DataArray):
+        """Query object for site location from site ID
+
+        Args:
+            long: DataArray of the longitude coordinates for any given site ID
+            lat: DataArray of the latitude coordinates for any given site ID
+
+        """
+        self.longitude = long
+        self.latitude = lat
+
+    def __call__(self, ml_id: int) -> Location:
+        """Returns the locations for the input GSP IDs.
+
+        Args:
+            gsp_id: Integer ID of the GSP
+        """
+        return Location(
+            coordinate_system="lon_lat",
+            x=self.longitude.sel(pv_system_id=ml_id).item(),
+            y=self.latitude.sel(pv_system_id=ml_id).item(),
+            id=ml_id,
+        )
 
 def extras(config: DictConfig) -> None:
     """A couple of optional utilities.
--- a/pvnet/utils.py
+++ b/pvnet/utils.py
@@ -68,7 +68,8 @@ class GSPLocationLookup:
             id=gsp_id,
         )
 
-class SiteLocationLookup:
+class SiteLocationLookup: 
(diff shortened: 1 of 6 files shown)


pvnet/utils.py
```diff
@@ -69,6 +69,34 @@ def __call__(self, gsp_id: int) -> Location:
         )
 
 
+class SiteLocationLookup:
```
Sukh-P, line 72: Not added unit tests for this as this is used only in a single place (site backtest script) so far and is similar to the class above for GSP location lookup (not quite similar enough though IMO that it's worth merging them into one class)

scripts/backtest_sites.py
```diff
@@ -0,0 +1,447 @@
...
+            ),
+        )
+
+        with torch.no_grad():
+            # Run batch through model to get 0-1 predictions for all sites
+            device_batch = copy_batch_to_device(batch_to_tensor(batch), device)
+            y_normed_site = model(device_batch).detach().cpu().numpy()
+        da_normed_site = preds_to_dataarray(y_normed_site, model, valid_times, ALL_SITE_IDS)
+
+        # TODO fix this step: Multiply normalised forecasts by capacities and clip negatives
```
peterdudfield, line 331: Whats wrong with it?

pvnet/utils.py
peterdudfield: would it be easy to add a quick test, just incase? Makes it a bit more reobust for the future

scripts/backtest_sites.py
Sukh-P: Seemed to be mixing up which capacities were for which site, haven't dug into this fully yet but might just be a site datapipes issues, left the outputs as a normalised output for now which can be converted to MW in post processing

scripts/backtest_sites.py
```diff
@@ -0,0 +1,447 @@
+"""
```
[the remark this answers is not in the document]
Sukh-P: Multiple sites, you give it a list of site ids and it will generate outputs for those for the specific period specified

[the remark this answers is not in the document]
peterdudfield: could you add that to the description, or something like that

scripts/backtest_sites.py
peterdudfield: thanks

scripts/backtest_sites.py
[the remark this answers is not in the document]
Sukh-P: Done

peterdudfield (approved):
Looks good, just remember to get that precommit step working. 
Normally its a linting error

pvnet/utils.py
Sukh-P: I've now added a parametrised unit test for this

Issue #168, opened 2024-03-26
Not all model config values logged accurately in weights & bias UI


Sukh-P:
## Describe the bug
Currently when using W&B as the logger when training a model using PVNet not all config parameters seem to show accurately in the W&B UI, some show as `null` even when they are non null. 

See example screenshot:
<img width="725" alt="Screenshot 2024-03-26 at 10 42 39" src="https://github.com/openclimatefix/PVNet/assets/42407101/d95d2ab1-109b-431d-b1e3-8f5618c83bb8">

## To Reproduce
Train a model using PVNet and use W&B as the logger and look at the config pane on the overview page of the run and compare the config that was actually used (should be saved in a local file when running PVNet). 

## Expected behaviour
Would expect W&B to log all the model config accurately so that comparisons can be made between different runs with different configs (not that currently a local config parameter file is produced so this information is not lost but would be favourable  

## Additional context
This seems to only happen for section of the config which have [formats such as this in the multimodal.yaml file](https://github.com/openclimatefix/PVNet/blob/main/configs.example/model/multimodal.yaml#L96), where `_target_` is the first key on the second level. Most likely is an issue with W&B parsing config in this specific format

peterdudfield:
Good to note its not a `wandb` issue as this works
```
import wandb

wandb_logger = dict(_target_="lightning.pytorch.loggers.wandb.WandbLogger",
                    project="wandb_test",
                    name="test_mode",
                    save_dir="wandb_test",
                    offline=False,  # set True to store all logs only locally
                    log_model=False,
                    job_type="train",
                    optimizer = dict(_target_="torch.optim.Adam", lr=0.1)
                    )

experiment = wandb.init(config=wandb_logger)
experiment.log({"loss": 0.1})
```

peterdudfield:
Also doesnt seem to be a problem with pytorch lightning
```
logger = WandbLogger(
    project="wandb_test",
    name="test_mode",
    save_dir="wandb_test",
    offline=False,  # set True to store all logs only locally
    log_model=False,
    job_type="train",
)
logger.log_hyperparams({"optimmizer":{"_target_":"torch.optim.Adam","lr": 0.1}})
logger.experiment.log({ "loss": 0.1})
```
works, so my guess is its a hydra issue

peterdudfield:
It looks like its happens in 
```
OmegaConf.save(config.model, f"{callback.dirpath}/model_config.yaml")
```
in pvnet/training.py

[...]

PR #217, opened 2024-06-11
save configuration to wandb

peterdudfield:
# Pull Request

## Description

- save configruation, 
- hydra and wandb should save it, but there is an issue with `_target_` dicts, and tricky to find out where the error was. Easier to save the file directly

#168 

## How Has This Been Tested?

Ran it locally

- [ ] Yes

## Checklist:

### DIFF
--- a/pvnet/training.py
+++ b/pvnet/training.py
@@ -124,7 +124,9 @@ def train(config: DictConfig) -> Optional[float]:
 
                 # upload configuration up to wandb
                 OmegaConf.save(config, "./experiment_config.yaml")
-                wandb_logger.experiment.save(f"{callback.dirpath}/data_config.yaml", callback.dirpath)
+                wandb_logger.experiment.save(
+                    f"{callback.dirpath}/data_config.yaml", callback.dirpath
+                )
                 wandb_logger.experiment.save("./experiment_config.yaml")
 
                 break
--- a/pvnet/training.py
+++ b/pvnet/training.py
@@ -123,11 +123,9 @@ def train(config: DictConfig) -> Optional[float]:
                 shutil.copyfile(data_config, f"{callback.dirpath}/data_config.yaml")
 
                 # upload configuration up to wandb
-                os.makedirs("./configuration", exist_ok=True)
-                shutil.copyfile(data_config, f"./configuration/data_config.yaml")
-                OmegaConf.save(config, f"./configuration/config.yaml")
-                wandb_logger.experiment.save(f"./configuration/data_config.yaml")
-                wandb_logger.experiment.save(f"./configuration/config.yaml")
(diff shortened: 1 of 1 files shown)


pvnet/training.py
```diff
@@ -121,6 +121,14 @@ def train(config: DictConfig) -> Optional[float]:
 
                 assert os.path.isfile(data_config), f"Data config file not found: {data_config}"
                 shutil.copyfile(data_config, f"{callback.dirpath}/data_config.yaml")
+
+                # upload configuration up to wandb
+                os.makedirs("./configuration", exist_ok=True)
+                shutil.copyfile(data_config, "./configuration/data_config.yaml")
+                OmegaConf.save(config, "./configuration/config.yaml")
+                wandb_logger.experiment.save("./configuration/data_config.yaml")
```
dfulu, line 129: Could we not just upload `f"{callback.dirpath}/model_config.yaml"` and `f"{config.datamodule.batch_dir}/data_configuration.yaml"` since these already exist rather than saving them to `./configuration` first?

dfulu (changes requested):

pvnet/training.py
peterdudfield: The file structure in wandb matches the file strcutre of the files uploaded, and I wanted it neat, rather than some `mnt/storage4tb_b/..../` for the data configuration. But perhaps there is a neater way to do it?

dfulu: I tried running this:
```
logger.experiment.save("/home/jamesfulton/repos/PVNet_summation/checkpoints/znoi18se/model_config.yaml")
```
and the file appears [here](https://wandb.ai/openclimatefix/pvnet_summation/runs/y745prto/files?nw=nwuserjames_ocf) without any subdirectories. So that suggests `logger.experiment.save(f"{callback.dirpath}/model_config.yaml")` should work and save without subdirs? 

But maybe I'm missing something, and if that doesn't work can we just delete `./configuration` after uploading to wandb to keep things tidy?

peterdudfield: hmm, yea, ok I could try that. 
How do you feel about override the current `config.yaml` in wandb with a new one? My feeling is that we should save both

dfulu: Yeh I'd save both

peterdudfield: It looks like I have to use `wandb_logger.experiment.save(f"{callback.dirpath}/data_config.yaml", callback.dirpath)` to make sure its not in a sub folder. Im on a slightly high wandb version that you @dfulu

peterdudfield: For example https://wandb.ai/openclimatefix/wandb_test/runs/b98z16i9/files?nw=nwuserpeterdudfield

dfulu (approved):

PR #224, opened 2024-06-18
Remove old experiment commands

dfulu:
Remove the the old file storing the commands for different experiments. These commands are now out of date with the current usage and the usage is shown in the README now anyway


Sukh-P (approved):
Good shout, btw I think the unrelated tests which are failing are due to numpy 2.0.0 being used in ocf_datapipes

PR #236, opened 2024-07-24
Use concurrent batch pipeline for ~30x speed up

dfulu:
Add the PVNet concurrent datapipe to the concurrent batch creation and backtest scripts. The new data is already in `ocf_datapipes` and is about 30x faster than the old datapipe for creating concurrent batches

### DIFF
--- a/pvnet/models/base_model.py
+++ b/pvnet/models/base_model.py
@@ -94,7 +94,7 @@ def minimize_data_config(input_path, output_path, model):
         if not model.include_nwp:
             del config["input_data"]["nwp"]
         else:
-            for nwp_source in list(config["input_data"]["nwp"].keys()):
+            for nwp_source in config["input_data"]["nwp"].keys():
                 nwp_config = config["input_data"]["nwp"][nwp_source]
 
                 if nwp_source not in model.nwp_encoders_dict:
--- a/pvnet/models/base_model.py
+++ b/pvnet/models/base_model.py
@@ -94,7 +94,7 @@ def minimize_data_config(input_path, output_path, model):
         if not model.include_nwp:
             del config["input_data"]["nwp"]
         else:
-            for nwp_source in config["input_data"]["nwp"].keys():
+            for nwp_source in list(config["input_data"]["nwp"].keys()):
                 nwp_config = config["input_data"]["nwp"][nwp_source]
 
                 if nwp_source not in model.nwp_encoders_dict:
--- a/scripts/backtest_uk_gsp.py
+++ b/scripts/backtest_uk_gsp.py
@@ -139,7 +139,7 @@ def get_available_t0_times(start_datetime, end_datetime, config_path):
     # Pop out the config file
     config = datapipes_dict.pop("config")
 
-    # We are going to abuse the `create_datapipes()` function to find the init-times in
+    # We are going to abuse the `create_t0_datapipe()` function to find the init-times in
     # potential_init_times which we have input data for. To do this, we will feed in some fake GSP
     # data which has the potential_init_times as timestamps. This is a bit hacky but works for now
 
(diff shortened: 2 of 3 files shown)


scripts/backtest_uk_gsp.py
```diff
@@ -144,7 +140,7 @@ def get_available_t0_times(start_datetime, end_datetime, config_path):
     # Pop out the config file
     config = datapipes_dict.pop("config")
 
-    # We are going to abuse the `create_t0_and_loc_datapipes()` function to find the init-times in
+    # We are going to abuse the `create_datapipes()` function to find the init-times in
```
zakwatts, line 143: genuine questions, is "abuse" a technical term here?

zakwatts: or does it mean my interpretation which is "use in a hacky way"

dfulu: No its not. But we are using this function in a really dirty way. We should really be writing a new clean function which does exactly what we want and refactoring create_datapipes rather than what we do here

zakwatts:
Looks good!

zakwatts (approved):

zakwatts:
Might be worth getting @AUdaltsova and @Sukh-P to review also as they use datapipes quite a bit

pvnet/models/base_model.py
```diff
@@ -94,7 +94,7 @@ def minimize_data_config(input_path, output_path, model):
         if not model.include_nwp:
             del config["input_data"]["nwp"]
         else:
-            for nwp_source in config["input_data"]["nwp"].keys():
+            for nwp_source in list(config["input_data"]["nwp"].keys()):
```
Sukh-P, line 97: Out of interest why has list been added? Is there a difference between looping through the items in the list vs the view object or something else?

scripts/backtest_uk_gsp.py
```diff
@@ -193,21 +186,19 @@ def get_available_t0_times(start_datetime, end_datetime, config_path):
...
-        tuple: A tuple of datapipes
-            - Datapipe yielding locations
-            - Datapipe yielding init-times
+        Datapipe: A Datapipe yielding init-times
     """
 
     # Set up ID location query object
     ds_gsp = get_gsp_ds(config_path)
-    gsp_id_to_loc = GSPLocationLookup(ds_gsp.x_osgb, ds_gsp.y_osgb)
+    GSPLocationLookup(ds_gsp.x_osgb, ds_gsp.y_osgb)
```
Sukh-P, line 201: I might be missing something here, is this missing an assignment to a variable or does it do something else? Can't see it referenced elsewhere

scripts/backtest_uk_gsp.py
```diff
@@ -173,17 +169,14 @@ def get_available_t0_times(start_datetime, end_datetime, config_path):
     datapipes_dict["gsp"] = IterableWrapper([ds_fake_gsp])
 
     # Use create_t0_and_loc_datapipes to get datapipe of init-times
```
Sukh-P, line 171: A nit but I guess this comment is now slightly out of date?

Sukh-P:
Great work adding this in and getting the speedup! Just checking has this been reran and tested locally?

pvnet/models/base_model.py
dfulu: Yeh good point. I had that change locally for some reason I've forgotten. I'm pretty sure you're right, so I'll change back

scripts/backtest_uk_gsp.py
dfulu: Nope, you're right. That was just a blunder

scripts/backtest_uk_gsp.py
dfulu: Yep, thank you!

dfulu:
> Great work adding this in and getting the speedup! Just checking has this been reran and tested locally?

Yeh, it has been tested locally on a small set of examples. I'm going to leave it open for now since I am making concurrent batches to train on and Zak is doing a backtest. Just in case more bugs come out of the woodwork during those

Sukh-P (approved):

Issue #235, opened 2024-07-24
Need a separate model card template for HF for windnet


AUdaltsova:
Currently we have just one template for model card when pushing to hugging face and it's for PVNet2. Might want to add a new one for WindNet

## Possible Implementation
The template is here `pvnet/models/model_card_template.md`, so would probably want to rename it to something like `model_card_template_pvnet.md` and create an updated copy in the same directory for windnet under `model_card_template_windnet.md`

Can even do an automated selection in `scripts/checkpoint_to_huggingface.py` based on HF repo name parsing, something like 
```
model = hf_repo_name.split('_')[0]
if model == "windnet": model_card = "PATH/model_card_template_windnet.md"
else:  model_card = "PATH/model_card_template_pvnet.md"
```

PR #239, opened 2024-07-24
Update default for wandb_ids

Sukh-P:
# Pull Request

## Description

Have run into this a couple times now where I have ran the script with just the `checkpoint_dir_path` parameter and I got an error about the `wandb_ids` being None, I realised it didn't go into the logic to get the run ID starting at line 43 since that checks if it is an empty list and the default is None and then runs into an error if it's not an ensemble due to

```
if not is_ensemble:
        wandb_ids = wandb_ids[0]
```

so I think changing it's default to an empty list might work a little more smoothly

### DIFF
--- a/scripts/checkpoint_to_huggingface.py
+++ b/scripts/checkpoint_to_huggingface.py
@@ -21,7 +21,7 @@ huggingface_repo = "openclimatefix/pvnet_uk_region"
 def push_to_huggingface(
     checkpoint_dir_paths: list[str],
     val_best: bool = True,
-    wandb_ids: Optional[list[str]] = None,
+    wandb_ids: list[str] = [],
     local_path: Optional[str] = None,
     push_to_hub: bool = True,
 ):


scripts/checkpoint_to_huggingface.py
```diff
@@ -21,7 +21,7 @@
 def push_to_huggingface(
     checkpoint_dir_paths: list[str],
     val_best: bool = True,
-    wandb_ids: Optional[list[str]] = None,
+    wandb_ids: list[str] = [],
```
AUdaltsova, line 24: Out of curiosity: why not Optional anymore?

AUdaltsova (approved):

scripts/checkpoint_to_huggingface.py
Sukh-P: That was a mistake, added it back in, thanks! Also updated the type hint slightly to the more up to date python way so we don't need to use `Optional`

dfulu:
This is a partial fix and it might be worth implementing a more complete one or making an issue to do so. 

The script adds [the link to the wandb in the huggingface model card](https://huggingface.co/openclimatefix/pvnet_uk_region#results). For this it uses a [hard-coded wandb project and hf repo](https://github.com/openclimatefix/PVNet/blob/main/scripts/checkpoint_to_huggingface.py#L17-L18) since it was written when we only used PVNet for UK regional. Having that automatically generated link between the model on HF and the actual training run is really good for traceability of models for us, and is also good for our open source-ness. 

I presume that you must be changing some of these hard coded paths locally in order to run this script anyway?

dfulu:
I've just seen that the [windnet](https://huggingface.co/openclimatefix/windnet_india#results) and [pvnet-india](https://huggingface.co/openclimatefix/pvnet_india#results) HF repos have used the model cards from PVNet UK which mention the model being used for UK regional GSP predictions. Also the links to the training runs lead to nowhere because of these hard coded values.

So I think this should probably be a separate issue. Misusing the model cards like this is a bit messy of us

Sukh-P:
> This is a partial fix and it might be worth implementing a more complete one or making an issue to do so.
> 
> The script adds [the link to the wandb run in the huggingface model card](https://huggingface.co/openclimatefix/pvnet_uk_region#results). For this it uses a [hard-coded wandb project and hf repo](https://github.com/openclimatefix/PVNet/blob/main/scripts/checkpoint_to_huggingface.py#L17-L18) since it was written when we only used PVNet for UK regional. Having that automatically generated link between the model on HF and the actual training run is really good for traceability of models for us, and is also good for our open source-ness.
> 
> I presume that you must be changing some of these hard coded paths locally in order to run this script anyway?

Yep so we edit these locally when running the script as it wouldn't be put in the correct HF repo otherwise (for the HF one), I guess this could be done more clearly and these be included as required parameters in the function itself to reduce the chance of some forgetting to change these

Sukh-P:
> I've just seen that the [windnet](https://huggingface.co/openclimatefix/windnet_india#results) and [pvnet-india](https://huggingface.co/openclimatefix/pvnet_india#results) HF repos have used the model cards from PVNet UK which mention the model being used for UK regional GSP predictions. Also the links to the training runs lead to nowhere because of these hard coded values.
> 
> So I think this should probably be a separate issue. Misusing the model cards like this is a bit messy of us

Yep we agree, @AUdaltsova  and I noticed this too and Alex has created an issue here for this https://github.com/openclimatefix/PVNet/issues/235

AUdaltsova:
> Yep so we edit these locally when running the script as it wouldn't be put in the correct HF repo otherwise (for the HF one), I guess this could be done more cleanly and these be included as required parameters in the function itself to reduce the chance of some forgetting to change these

Can we maybe add choices/hints for it? Like "must be one of: path-to-pvnet-repo, path-to-windnet-repo" etc or even just in comments somewhere so you don't have to go look up the specific name on HF? Not that it's a lot of work but I'm a big fan of human laziness accommodation

Sukh-P:
FYI I am trying to tackle #235 in this PR too just to try and clean this up a bit more in one go

pvnet/models/model_cards/pvnet_india_model_card_template.md
```diff
@@ -0,0 +1,51 @@
...
+
+
+
+
+# PVNet India
+
+## Model Description
+
+<!-- Provide a longer summary of what this model is/does. -->
+This model class uses numerical weather prediction and recent wind power output to forecast the PV power in North West India over the next 48 hours. More information can be found in the model repo [1] and experimental notes [here](https://github.com/openclimatefix/PVNet/tree/main/experiments/india).
```
AUdaltsova, line 15: I'm nitpicking into the sunset here, but I think technically only solar power is PV power. Unless I'm misunderstanding what it says of course!

Sukh-P:
Okay should be good to go now for a review with the additional changes, thanks

pvnet/models/base_model.py
```diff
@@ -277,19 +281,28 @@ def save_pretrained(
...
 
+        # Get appropriate model card
+        model_name = huggingface_repo.split("/")[1]
+        if model_name == "windnet_india":
+            model_card = "windnet_model_card_template.md"
+        elif model_name == "pvnet_india":
+            model_card = "pvnet_india_model_card_template.md"
+        else:
+            model_card = "model_card_template.md"
+
```
AUdaltsova, line 292: hurray! very nice :)

AUdaltsova (approved):
Looks good to me!

pvnet/models/model_cards/pvnet_india_model_card_template.md
AUdaltsova, line 15: Out of curiosity: pvsite really uses wind too?

pvnet/models/base_model.py
dfulu: Can we rename `model_card_template.md` to something like `uk_regional_model_card_template.md` and `pvnet_india_model_card_template.md` to something like `india_[pv? wind?]_model_card_template.md`? I think we should be very explicit with the names

pvnet/models/model_cards/pvnet_india_model_card_template.md
Sukh-P: Woops my bad

pvnet/models/base_model.py
Sukh-P: Yep good shout, let me change that

dfulu (approved):
Looks really good!

PR #241, opened 2024-08-05
Fix history bug and add bells in analysis.py

AUdaltsova:
# Pull Request

## Description
Analysis.py results were inconsistent because `run.history()` was pulling 500 random steps instead of the whole history. Fixed that.

Bells and whistles:
- Added epoch number to the legend label for easier assessment. May be redundant, happy to remove it

## How Has This Been Tested?
Ran analysis.py several times for the same set of runs to make sure it gave consistent results; separately tested that run.lastHistoryStep was working as expected

- [X] Yes

## Checklist:

### DIFF
--- a/experiments/analysis.py
+++ b/experiments/analysis.py
@@ -16,10 +16,11 @@ def main(runs: list[str], run_names: list[str]) -> None:
     """
     api = wandb.Api()
     dfs = []
+    epoch_num = []
     for run in runs:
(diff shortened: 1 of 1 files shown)


AUdaltsova:
@peterdudfield turns out I made the change and never merged, mind taking a quick look when you have a minute?

peterdudfield (approved):

PR #252, opened 2024-09-05
save validation batch results to wandb

peterdudfield:
# Pull Request

## Description

Save csv of validation results to weights and biases

This is done by 
- results being appended each validation step,
- saved all together at the end of the validation step 

## How Has This Been Tested?

CI tests
- local test with script

```python
import wandb
import pandas as pd

run = wandb.init(
    project="test",
    notes="My first experiment",
)

# make random dataframe
df = pd.DataFrame({
    "a": [1, 2, 3],
    "b": [4, 5, 6]
})

i=2
validation_artifact = wandb.Artifact(f"validation_results_{i}", type="dataset")
df.to_csv(f"validation_results_{i}.csv", index=False)
validation_artifact.add_file(f"validation_results_{i}.csv")
wandb.log_artifact(validation_artifact)

```
 
Screenshot 2024-09-05 at 16 15 28

- [ ] Yes

## Checklist:

### DIFF
--- a/pvnet/models/base_model.py
+++ b/pvnet/models/base_model.py
@@ -635,7 +635,11 @@ class BaseModel(pl.LightningModule, PVNetModelHubMixin):
 
         for i in range(batch_size):
             y_i = y[i]
-            y_hat_i = y_hat[i]
+            if self.use_quantile_regression:
+                idx = self.output_quantiles.index(0.5)
+                y_hat_i = y_hat[i,idx]
+            else:
+                y_hat_i = y_hat[i]
             time_utc_i = time_utc[i]
             target_id_i = target_id[i]
 
--- a/pvnet/models/base_model.py
+++ b/pvnet/models/base_model.py
@@ -617,7 +617,7 @@ class BaseModel(pl.LightningModule, PVNetModelHubMixin):
         """Append validation results to self.validation_epoch_results"""
 
         # get truth values, shape (b, forecast_len)
-        y = batch[self._target_key][:, -self.forecast_len:, 0]
+        y = batch[self._target_key][:, -self.forecast_len :, 0]
         y = y.detach().cpu().numpy()
         batch_size = y.shape[0]
 
@@ -626,7 +626,7 @@ class BaseModel(pl.LightningModule, PVNetModelHubMixin):
 
         # get time_utc, shape (b, forecast_len)
         time_utc_key = BatchKey[f"{self._target_key_name}_time_utc"]
-        time_utc = batch[time_utc_key][:, -self.forecast_len:].detach().cpu().numpy()
+        time_utc = batch[time_utc_key][:, -self.forecast_len :].detach().cpu().numpy()
 
         # get target id and change from (b,1) to (b,)
         id_key = BatchKey[f"{self._target_key_name}_id"]
--- a/pvnet/models/base_model.py
+++ b/pvnet/models/base_model.py
@@ -623,18 +623,13 @@ class BaseModel(pl.LightningModule, PVNetModelHubMixin):
             y_i = y[i].detach().cpu().numpy()
             y_hat_i = y_hat[i].detach().cpu().numpy()
 
-            print(BatchKey._member_map_)
-            try:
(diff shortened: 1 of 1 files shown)


pvnet/models/base_model.py
```diff
@@ -675,6 +713,21 @@ def validation_step(self, batch: dict, batch_idx):
...
+            # join together validation results, and save to wandb
+            validation_results_df = pd.concat(self.validation_epoch_results)
+            with tempfile.TemporaryDirectory() as tempdir:
+                filename = os.path.join(tempdir, f"validation_results.csv_{self.current_epoch}")
+                validation_results_df.to_csv(filename, index=False)
+
+                validation_artifact = wandb.Artifact(
+                    f"validation_results_epoch={self.current_epoch}", type="dataset"
+                )
+                wandb.log_artifact(validation_artifact)
```
dfulu, line 726: Compared to your example, I think you're missing a line
```
validation_artifact = wandb.Artifact(f"validation_results_{i}", type="dataset")
df.to_csv(f"validation_results_{i}.csv", index=False)
validation_artifact.add_file(f"validation_results_{i}.csv")   <---- this one
wandb.log_artifact(validation_artifact)
```

pvnet/models/base_model.py
```diff
@@ -609,12 +613,48 @@ def _log_forecast_plot(self, batch, y_hat, accum_batch_num, timesteps_to_plot, p
...
+
+            id_key = BatchKey[f"{self._target_key_name}_id"]
+            target_id = batch[id_key][i].detach().cpu().numpy()
+            if target_id.ndim > 0:
+                target_id = target_id[0]
+
+            results_df = pd.DataFrame(
+                {
+                    "y": y_i,
+                    "y_hat": y_hat_i,
```
dfulu, line 637: Have you done a training run with this? When predicting quantiles, I think `y_i` and `y_hat_i` will be different shapes

pvnet/models/base_model.py
```diff
@@ -609,12 +613,48 @@ def _log_forecast_plot(self, batch, y_hat, accum_batch_num, timesteps_to_plot, p
...
         plt.close(fig)
 
+    def _log_validation_results(self, batch, y_hat, accum_batch_num):
+        """Append validation results to self.validation_epoch_results"""
+
+        y = batch[self._target_key][:, -self.forecast_len :, 0]
+        batch_size = y.shape[0]
+
+        for i in range(batch_size):
+            y_i = y[i].detach().cpu().numpy()
```
dfulu, line 623: I think at least some of this detaching and moving to CPU should happen before the loop starts. Doing it in the loop means the CPU and GPU need to communicate with each other a lot more

dfulu (changes requested):

pvnet/models/base_model.py
peterdudfield: Thanks good catch

pvnet/models/base_model.py
peterdudfield: I havent done training, but the end2end test does go through here. 

becasue i take `y = batch[self._target_key][:, -self.forecast_len :, 0]`, it makes it the same length as y_hat. 

Perhaps theres a better way to standardised that. something for the future

pvnet/models/base_model.py
peterdudfield: Thanks, good spot, yea ill move this out the loop

pvnet/models/base_model.py
dfulu: Yeh but I think in this case, `y_i` is a vector with shape `(horizon_step,)`. But `y_hat_i` can either be a vector with shape `(horizon_step,)` or `(horizon_step, quantile,)` depending on whether we are training to predict quantiles or only a central value. The end2end test only tests non-quantile training.

peterdudfield: oh good point. ill have a think about the qunalite things. Good catch on that

peterdudfield: I should have it now

dfulu (approved):

PR #256, opened 2024-09-25
add flushing of val epoch resluts

### LINKED ISSUE #255: wandb csv artifact saves too much data to local storage
## Describe the bug
Not exactly a bug but, when running offline wandb stores the new .csv files of inference to your .local and .cash folders, and that's quite a lot of data (I got 60G before my storage maxed out and broke training). Which is not exactly criminal but is quite a lot of space so we probably want to cut it down or at least be able to say "stop" somehow.

## Suggested fix
Assuming we can forgo having inference at every 50 steps, I'm looking into only keeping best epoch results and cleaning up all else; but all suggestions welcome
Or, I guess, we can just always run wandb online for immediate sync without storing locally, but it feels wrong

AUdaltsova:
# Pull Request

## Description

array that stores inference results for wandb csv saving was not getting cleaned between val epochs, resulting in blowing up file sizes. Added removing results after logging.

Potentially fixes #255 as the files will be kept to 19-20 MB (for me at least; obviously will be different for different configs, this is 3 quantiles and 32 points predictions) instead of getting to 240+MB by epoch 11 etc. (Still need to remove keeping multiple versions of one artifact, not sure how to do that yet)

## How Has This Been Tested?
Run training for 3 epochs, checked resulting csvs had expected sizes and contents

- [X] Yes

## Checklist:

### DIFF
--- a/pvnet/models/base_model.py
+++ b/pvnet/models/base_model.py
@@ -666,7 +666,8 @@ class BaseModel(pl.LightningModule, PVNetModelHubMixin):
         # Sensor seems to be in batch, station, time order
         y = batch[self._target_key][:, -self.forecast_len :, 0]
 
-        self._log_validation_results(batch, y_hat, accum_batch_num)
+        if (batch_idx + 1) % self.trainer.accumulate_grad_batches == 0:
+            self._log_validation_results(batch, y_hat, accum_batch_num)
 
         # Expand persistence to be the same shape as y
         losses = self._calculate_common_losses(y, y_hat)
--- a/pvnet/models/base_model.py
+++ b/pvnet/models/base_model.py
@@ -1,4 +1,3 @@
-"""Base model for all PVNet submodels"""
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


AUdaltsova:
Ok to be fair turns out versioning happens because different runs in one project save their artifacts under the same name, so one run has one file per epoch instead of multiple as I originally thought. The naming might become a problem though bc afaik the main usecase for these files is to be downloaded for further analysis? If so getting the correct one can get kinda tricky, maybe we want to prefix them by run or something. Or maybe people who can use wandb.api better than me can actually do it neatly and this is a non-isssue?

peterdudfield:
> Ok to be fair turns out versioning happens because different runs in one project save their artifacts under the same name, so one run has one file per epoch instead of multiple as I originally thought. The naming might become a problem though bc afaik the main usecase for these files is to be downloaded for further analysis? If so getting the correct one can get kinda tricky, maybe we want to prefix them by run or something. Or maybe people who can use wandb.api better than me can actually do it neatly and this is a non-isssue?

Weird, I would have thought each articfact is seperate for experiment, but maybe im wrong there. is there a model id we can use to suffix it?

Sukh-P (approved):

PR #259, opened 2024-10-07
Bug fixes

dfulu:
I encountered some bugs when pushing the latest version of the UK model to huggingface. These fixes were required for it to go through

- Type hinting changes due to limitations with typer
- Command line `_` to `-` 
- Change in looping over the dictionary keys since we can't loop over the keys in a dict whilst editing that dict


Sukh-P (approved):
Thanks for updating these!

PR #266, opened 2024-10-23
update backtest_sites.py

AUdaltsova:
# Pull Request

## Description

Decoupled pv padding from data so it uses config instead as had problems running without it and wrong padding (t0 index is not actually the true index when there is not enough data as it's calculated once with respect to proper batch shape and not for each sample)

## How Has This Been Tested?

Reran small backtest to check the results ar the same

- [X] Yes

## Checklist:


Sukh-P (approved):
LGTM!

PR #276, opened 2024-11-13
Upgrade UK to data sampler 

dfulu:
This PR upgrades the UK part datamodule to ocf-data-sampler. 

Notes:
- The datamodules for wind and pvsite have not been updated
- This breaks PVNet for wind and pvsites due to an update in the target variable shape coming from data-sampler
- I have also stripped out the exponential weighted losses. We haven't used them in years


pvnet/data/datamodule.py
```diff
@@ -1,57 +1,138 @@
...
-from torch.utils.data.datapipes.iter import FileLister
 
-from pvnet.data.base import BaseDataModule
+from ocf_datapipes.batch import batch_to_tensor, stack_np_examples_into_batch, NumpyBatch
+from ocf_data_sampler.torch_datasets.pvnet_uk_regional import (
+    PVNetUKRegionalDataset
+)
+
+
+def fill_nans_in_arrays(batch):
```
dfulu, line 15: This was put into the data-sampler in openclimatefix/ocf-data-sampler#49 so could be removed now

pvnet/models/multimodal/multimodal.py
```diff
@@ -365,7 +362,7 @@ def forward(self, x):
 
         # ********************** Embedding of GSP ID ********************
         if self.embedding_dim:
-            id = x[BatchKey[f"{self._target_key_name}_id"]][:, 0].int()
+            id = x[BatchKey[f"{self._target_key_name}_id"]][:].int()
```
dfulu, line 365: This should be made cleaner to remove the unnecessary `[:]`

pvnet/utils.py
```diff
@@ -265,14 +265,14 @@ def _get_numpy(key):
     y_id_key = BatchKey[f"{key_to_plot}_id"]
     BatchKey[f"{key_to_plot}_t0_idx"]
     time_utc_key = BatchKey[f"{key_to_plot}_time_utc"]
-    y = batch[y_key][:, :, 0].cpu().numpy()  # Select the one it is trained on
+    y = batch[y_key][:, :].cpu().numpy()  # Select the one it is trained on
```
dfulu, line 268: Remove unnecessary `[:,:]`

scripts/checkpoint_to_huggingface.py
```diff
@@ -19,10 +19,10 @@
 def push_to_huggingface(
     checkpoint_dir_paths: list[str],
     huggingface_repo: str = "openclimatefix/pvnet_uk_region",  # e.g. openclimatefix/windnet_india
-    wandb_repo: str | None = "openclimatefix/pvnet2.1",
```
dfulu, line 22: typer seems unable to work with types hints including a union like this

pvnet/data/datamodule.py
```diff
@@ -1,57 +1,111 @@
...
+        return len(self.sample_paths)
 
-rlimit = resource.getrlimit(resource.RLIMIT_NOFILE)
-resource.setrlimit(resource.RLIMIT_NOFILE, (2048, rlimit[1]))
+    def __getitem__(self, idx):
+        return torch.load(self.sample_paths[idx])
 
 
-class DataModule(BaseDataModule):
+def collate_fn(samples: list[NumpyBatch]):
```
Sukh-P, line 29: Small one but it might be nice to have a return type hint here:
```suggestion
def collate_fn(samples: list[NumpyBatch]) -> TensorBatch:
```
just so it's a bit clearer what this function returns, would require importing that type from ocf_datapipes and eventually we would put this or something similar into ocf-data-sampler to remove that dependency

dfulu: Yeh ,sounds good

tests/test_data/presaved_samples/data_configuration.yaml
```diff
@@ -0,0 +1,130 @@
...
+        - sr
+        - duvrs
+
+      nwp_image_size_pixels_height: 12 # roughly equivalent to ukv 48
+      nwp_image_size_pixels_width: 12 # roughly equivalent to ukv 48
+      dropout_timedeltas_minutes: [-360]
+      dropout_fraction: 1.0
+      max_staleness_minutes: null
+
+    sat_pred:
```
Sukh-P, line 81: Is the sat_pred section needed currently?

pvnet/models/base_model.py
```diff
@@ -582,7 +570,7 @@ def _training_accumulate_log(self, batch, batch_idx, losses, y_hat):
     def training_step(self, batch, batch_idx):
         """Run training step"""
         y_hat = self(batch)
-        y = batch[self._target_key][:, -self.forecast_len :, 0]
+        y = batch[self._target_key][:, -self.forecast_len :]
```
Sukh-P, line 573: Could you remind me what the shape change is here for the target? A comment could be helpful here on the shape

Sukh-P:
More of a general question, looking through am I right in thinking that we are (for the UK regional stuff at least) moving from saving batches (batch_size number of samples saved together) to saving each individual sample in separate .pt files, does the extra writes (a write to .pt for each sample now) and reads impact batch creation time / training times significantly or is it negligible?

pvnet/models/base_model.py
```diff
@@ -414,6 +412,10 @@ def __init__(
         # save all validation results to array, so we can save these to weights n biases
         self.validation_epoch_results = []
 
+    def transfer_batch_to_device(self, batch, device, dataloader_idx):
```
Sukh-P, line 415: I feel like I might be missing something here, where is this function used?

tests/test_data/presaved_samples/data_configuration.yaml
dfulu: This basically labels what is in the presaved samples which are there - i.e. this is the config which was used to produce them. I just copied some samples I had already rather than making dedicated ones. 

Its a bit excessive, but for now I personally think its okay. Once we do a bit more work in ocf-data-sampler I'd like to get rid of the presaved samples and test the save_sample script in the tests, and use samples it creates

pvnet/models/base_model.py
dfulu: Well its just that the new y values from data-sampler are only of shape [batch, time]. Previously they were [batch, time, 1] which seemed a little untidy

pvnet/models/base_model.py
dfulu: Its used behind the scenes by lightning, the same way that say the `training_step()` function is. We were previously using lightning's default batch transfer function, but this gives us more explicit control

Sukh-P (approved):
Really great work, thanks for adding this in!

Sukh-P:
Starting to use this code to test out the Site Torch Dataset pipeline, whilst doing that I thought it would be good to update the config example parameters in the [datamodule folder](https://github.com/openclimatefix/PVNet/tree/data_sampler/configs.example/datamodule) which mention "batch" to "samples" to align with the new `save_samples.py` script

peterdudfield:
is it good to merge this? Make sure @Sukh-P is working from that branch then?

dfulu:
Well there is still the example configs to update, but yeh then we could merge into the `dev-data-sampler` branch

PR #290, opened 2024-12-17
Add sample saving for Site Dataset + presaved SiteDataset/DataModule for loading presaved samples

Sukh-P:
# Pull Request
Tackles #301 

## Description

Inclues:
- Adding support for creating samples using the `ocf-data-sampler` `Site` torch dataset 
- Adding small torch Dataset for premade netcdf samples
- Refactor Datamodules so there is a base datamodule and then SiteDatamodule which inherits from that (will move the UK regional Datamodule to this structure as well but it needs some updating to do with Batchkeys first, all this can be a separate PR)
- Removing the datapipe specified datamodules which won't be used after moving to `ocf-data-sampler`
- Small documentation updates in the README (may have missed some)

## How Has This Been Tested?

Both the save_samples script and the loading premade samples have been tested locally (further changes required to actually get model training fully without errors) 

Also added tests back in with new SiteDatamodule. 

**TODO Testing**

- Lots of the test and test data only work at the moment with the old BatchKey format, need to update them


pyproject.toml
```diff
@@ -6,7 +6,7 @@ dynamic = ["version", "readme"]
 license={file="LICENCE"}
 
 dependencies = [
-    "ocf_data_sampler==0.0.26",
+    "ocf_data_sampler==0.0.32",
```
Sukh-P, line 9: This needs updating to a new version after https://github.com/openclimatefix/ocf-data-sampler/pull/82/files is merged in, should fix the failing test

pvnet/data/datamodule.py
```diff
@@ -93,7 +93,14 @@ def __init__(
         )
 
     def _get_streamed_samples_dataset(self, start_time, end_time) -> Dataset:
-        return PVNetUKRegionalDataset(self.configuration, start_time=start_time, end_time=end_time)
+        if self.configuration.renewable == "pv":
```
peterdudfield, line 96: should we call this uk_pv, just to be really clear

Sukh-P: Yep that sounds good, we only need two values of this now I think, "uk_pv" and "site" really, perhaps we should make that clearer in documentation, unless the renewable parameter config is used somewhere else

tests/conftest.py
```diff
@@ -159,20 +158,21 @@ def sample_pv_batch():
...
-        num_workers=0,
-        prefetch_factor=None,
-        train_period=[None, None],
-        val_period=[None, None],
-        test_period=[None, None],
-        batch_dir="tests/test_data/sample_wind_batches",
-    )
-    batch = next(iter(dm.train_dataloader()))
-    return batch
+# TODO update this test once we add the loading logic for the Site dataset
```
peterdudfield, line 161: I cant work out if we should merge before doing this, im not sure there. Relates to https://github.com/openclimatefix/ocf-data-sampler/issues/99

peterdudfield (changes requested):
Code looks all good. I wonder if we should od https://github.com/openclimatefix/ocf-data-sampler/issues/99 first, then add some test cases in here, just to be double sure

pyproject.toml
```diff
@@ -6,7 +6,7 @@ dynamic = ["version", "readme"]
 license={file="LICENCE"}
 
 dependencies = [
-    "ocf_data_sampler==0.0.32",
+    "ocf_data_sampler==0.0.43",
```
Sukh-P, line 9: Note this update in `ocf-data-sampler` version will break things from working properly on this dev branch for a bit before we get to fixing/updating everything.

I see two options:

1. We don't merge this in until we get everything working, pros is that we don't need another branch, con is this PR will get massive (already fairly big) 
2. Perhaps can use another dev branch, the original is kept as the working dev branch for ocf-data-sampler in PVNet and the new one is the one which is the interim one whilst we get all things working. Pros and cons basically the opposite of 1. 

The reason for the caution is if we were to merge this in as is I think we lose any remote code which allows ocf-data-sampler to work with PVNet which feels risky? 

@peterdudfield @dfulu let me know what you think, may be other options too

peterdudfield: Hmm, yea I can see the decision here. Get the balance right between small PRs, and getting things working properly. I wouldnt commmit to `main`, but i know this is only `dev`. 

What is there to do in this bit of work to make sure it is working? Do you think it is just updating from BatchKeys to SampleKeys, a bit more? 

Without knowing this: I would probably still merge, and then fix afterwards. Its not the main branch, and people can pull a different commit if they need to. It also allows someone to fix issue X, why you doing issue Y.

Sukh-P: Yeah it's mainly moving away from the old `BatchKey` enum to the `SampleKey`s we use now I believe, it's used in a few places in PVNet

peterdudfield: yea, i think make a github issue with these things, and merge this one

Sukh-P: Are you happy to approve?

peterdudfield (approved):

dfulu (approved):

PR #302, opened 2025-01-21
Update PVNet for datasampler

Sukh-P:
# Pull Request

## Description

This PR includes:
- Updates PVNet to use the new batch format that ocf-data-sampler produces
- Updates some of the model modules (such as site encoder) to work with samples from the sites data pipeline. 
- Some test data has been updated as well to reflect the new batch format 

Some updates have not been made in this PR and may be better suited to a separate PR to separate concerns and reduce PR size:

- Using the new SampleKeys rather than strings in places (TBD whether that is what we want)
- Updating some of the network modules to the new batch format that aren't used as often (e.g one of the site encoders)
- Naming consistency, use `pv_encoder` in some places, I have left this in for now because there are some tests which rely on this (would need to update some HF files to get some from_pretrained tests to work) 

## How Has This Been Tested?

Getting tests to pass & running training on some premade site samples.


pvnet/utils.py
```diff
@@ -259,10 +258,9 @@ def plot_batch_forecasts(
     def _get_numpy(key):
         return batch[key].cpu().numpy().squeeze()
 
-    y_key = BatchKey[f"{key_to_plot}"]
-    y_id_key = BatchKey[f"{key_to_plot}_id"]
-    BatchKey[f"{key_to_plot}_t0_idx"]
-    time_utc_key = BatchKey[f"{key_to_plot}_time_utc"]
+    y_key = f"{key_to_plot}"
```
dfulu, line 261: Can we clean this up to remove the unneeded f-string

tests/test_data/presaved_samples_site/data_configuration.yaml
dfulu, line 1: Can we remove the commented out bits

pvnet/models/multimodal/multimodal.py
```diff
@@ -229,27 +221,14 @@ def __init__(
             assert pv_history_minutes is not None
 
             self.pv_encoder = pv_encoder(
-                sequence_length=pv_history_minutes // pv_interval_minutes + 1,
+                sequence_length=pv_history_minutes // pv_interval_minutes - 1,
```
dfulu, line 224: What is the reason to change this to a minus?

pvnet/models/multimodal/site_encoders/encoders.py
```diff
@@ -206,9 +206,11 @@ def __init__(
         )
 
     def _encode_inputs(self, x):
-        # Shape: [batch size, sequence length, PV site] -> [8, 197, 1]
+        # Shape: [batch size, sequence length, number of sites] -> [8, 197, 1]
```
dfulu, line 209: Whilst we're here. Those hard coded numbers in the comment aren't generally true right? We should strip them out

tests/conftest.py
```diff
@@ -159,16 +161,15 @@ def sample_pv_batch():
 
 
 @pytest.fixture()
-def sample_wind_batch():
+def sample_site_batch():
     dm = SiteDataModule(
         configuration=None,
-        batch_size=2,
+        batch_size=8,
```
dfulu, line 167: Is there a reason to use 8? This seems like it would just make the tests take longer

dfulu (changes requested):
Looks really good. Thanks for your thoroughness in these changes. I've just got a few questions and comments on here

pvnet/models/multimodal/multimodal.py
Sukh-P: Ah forgot to leave a comment here that this is a TODO to check why I needed to do this to get things to work, wanted to double check that the site time slicing is as expected because when I used an interval start of -60 and then there was some just using history logic in PVNet I got 3 data points of generation (15 minutely data) I was expecting 5, but I am going to go back and understand why that is and will revert this if needed, thanks for the reminder!

pvnet/models/multimodal/site_encoders/encoders.py
Sukh-P: Yes the numbers themselves aren't accurate, I can remove them!

tests/conftest.py
Sukh-P: I think I was just playing around with a few different numbers for testing so the 8 is arbitrary, can put back to 2 as I agree will speed up the tests a bit

pvnet/models/multimodal/multimodal.py
dfulu: Hmm okay. Does that mean the issue come from data-sampler?

Sukh-P: Unsure, will take a proper look once I have fixed merge conflicts and report back!

dfulu: Thanks!

Sukh-P: Got to the bottom of it, that shouldn't have been changed was an issue in the config I was using for local testing so I have reverted it back to what it was, thanks!

dfulu: Great! Thanks!

dfulu (approved):
Looks great @Sukh-P. Can we make it a squash merge? We haven't in the past in this repo but I think it is much cleaner

Issue #301, opened 2024-12-20
Add site training torchdataset


peterdudfield:
We need a torch dataset, that loads premade batches. 
This could live in PVnet, but I think it makes sense to live in this repo

The idea is to create a troch dataset that loads netcdf samples in, and change them to torch tensor, ready to be used by PVNet

Credit to @Sukh-P for this suggestion
```
class NetCDFPremadeSamplesDataset(Dataset):
    """Dataset to load NumpyBatch samples"""

    def __init__(self, sample_dir):
        """Dataset to load NumpyBatch samples

        Args:
            sample_dir: Path to the directory of pre-saved samples.
        """
        self.sample_paths = glob(f"{sample_dir}/*.nc")

    def __len__(self):
        return len(self.sample_paths)

    def __getitem__(self, idx):
        ds = xr.open_dataset(self.sample_paths[idx])
        return self.convert_to_numpy_batch(ds)
    
    def convert_to_numpy_batch(ds: xr.Dataset) -> NumpyBatch:
        da_dict = uncombine_from_single_dataset(ds) # this function is from ocf_datapipes
        numpy_batch = dict_of_arrays_to_numpy_batch(da_dict) # this function would need to be added in ocf-data-sampler, it would look similar to the process and combine except it would not need the normalisation steps
        return numpy_batch
```

And then in PVnet, we would have to update it to something like this
```
def _get_premade_samples_dataset(self, subdir) -> Dataset:
        split_dir = f"{self.sample_dir}/{subdir}"
        return NumpybatchPremadeSamplesDataset(split_dir)
```

Sukh-P:
This was added in this commit https://github.com/openclimatefix/PVNet/commit/075ae315c93c98ff345561603e9e834c8a1218e4

PR #311, opened 2025-01-29
Incorporate SiteSample and UKRegionalSample into save/load

felix-e-h-p:
save_samples.py, site_datamodule.py and uk_regional_datamodule.py updated accordingly to utilise new SiteSample and UKRegionalSample save/load functionality.

Additional test added for clarity and version update stated in dependencies.

### DIFF
--- a/pvnet/data/site_datamodule.py
+++ b/pvnet/data/site_datamodule.py
@@ -1,11 +1,9 @@
 """ Data module for pytorch lightning """
 
-import xarray as xr
+from ocf_data_sampler.sample.site import SiteSample
 from ocf_data_sampler.torch_datasets.datasets.site import (
     SitesDataset,
-    convert_netcdf_to_numpy_sample,
 )
-from ocf_data_sampler.sample.site import SiteSample
 from torch.utils.data import Dataset
 
 from pvnet.data.base_datamodule import BaseDataModule, PremadeSamplesDataset
--- a/pvnet/data/site_datamodule.py
+++ b/pvnet/data/site_datamodule.py
@@ -6,6 +6,7 @@ from ocf_data_sampler.torch_datasets.datasets.site import (
     SitesDataset,
     convert_netcdf_to_numpy_sample,
 )
+from ocf_data_sampler.sample.site import SiteSample
 from torch.utils.data import Dataset
 
 from pvnet.data.base_datamodule import BaseDataModule
@@ -29,12 +30,8 @@ class NetcdfPreMadeSamplesDataset(Dataset):
         return len(self.sample_paths)
 
     def __getitem__(self, idx):
-        # open the sample
-        ds = xr.open_dataset(self.sample_paths[idx])
-
-        # convert to numpy
-        sample = convert_netcdf_to_numpy_sample(ds)
-        return sample
+        sample = SiteSample.load(self.sample_paths[idx])        
+        return sample.to_numpy()
 
 
 class SiteDataModule(BaseDataModule):
--- a/pvnet/data/site_datamodule.py
+++ b/pvnet/data/site_datamodule.py
@@ -1,35 +1,14 @@
 """ Data module for pytorch lightning """
-from glob import glob
 
-from ocf_data_sampler.sample.site import SiteSample
+import xarray as xr
 from ocf_data_sampler.torch_datasets.datasets.site import (
     SitesDataset,
+    convert_netcdf_to_numpy_sample,
 )
+from ocf_data_sampler.sample.site import SiteSample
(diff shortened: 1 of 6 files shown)


pvnet/data/site_datamodule.py
```diff
@@ -29,12 +28,8 @@ def __len__(self):
...
 
     def __getitem__(self, idx):
-        # open the sample
-        ds = xr.open_dataset(self.sample_paths[idx])
-
-        # convert to numpy
-        sample = convert_netcdf_to_numpy_sample(ds)
-        return sample
+        sample = SiteSample.load(self.sample_paths[idx])
+        return sample.to_numpy()
```
peterdudfield, line 32: perfect

pvnet/data/uk_regional_datamodule.py
```diff
@@ -23,7 +23,8 @@ def __len__(self):
         return len(self.sample_paths)
 
     def __getitem__(self, idx):
-        return torch.load(self.sample_paths[idx])
+        sample = UKRegionalSample.load(self.sample_paths[idx])
+        return sample.to_numpy()
```
peterdudfield, line 27: maybe just add a comment that this returns a dictionary of tensors

peterdudfield (approved):
See comment that you could add for extra clarity

pvnet/data/uk_regional_datamodule.py
felix-e-h-p: Done, cheers!

Sukh-P:
So I reckon as @dfulu had previously suggested here https://github.com/openclimatefix/PVNet/issues/301#issuecomment-2592748362 for the loading and to_numpy bits this can be abstracted a bit more to essentially one `PremadeSamplesDataset` as the comment showed and that could live in the base_datamodule file and then in [these bits](https://github.com/openclimatefix/PVNet/blob/dev-data-sampler/pvnet/data/site_datamodule.py#L83) for the site and uk_regional data module you just need to use that `PremadeSamplesDataset` and pass in the correct sample class, either the site one or the uk_regional one

scripts/save_samples.py
```diff
@@ -76,10 +77,18 @@ def __init__(self, save_dir: str, renewable: str = "pv_uk"):
...
+            uk_sample._data = sample
+            uk_sample.save(f"{save_path}.pt")
         elif self.renewable == "site":
-            sample.to_netcdf(f"{self.save_dir}/{sample_num:08}.nc", mode="w", engine="h5netcdf")
+            # Create Site Sample and save
+            site_sample = SiteSample()
+            site_sample._data = sample
+            site_sample.save(f"{save_path}.nc")
         else:
             raise ValueError(f"Unknown renewable: {self.renewable}")
```
Sukh-P, line 94: Could this bit be simplified a bit more? 

Could have something like this: 

```
if self.renewable == "pv_uk":
    sample_class = UKRegionalSample()
elif self.renewable == "site":
    sample_class = SiteSample()
else:
    raise ValueError(f"Unknown renewable: {self.renewable}")
#  assign data and save
sample_class._data = sample
sample_class.save(f"{save_path}.pt")
```

Sukh-P (changes requested):
Thanks for adding these changes in, I think just needs a bit more refactoring to get the full benefit of having the generalised interface

scripts/save_samples.py
peterdudfield: if self.renewable == "pv_uk":
    sample_class = UKRegionalSample()
   filename = f"{save_path}.pt"
elif self.renewable == "site":
    sample_class = SiteSample()
     filename = f"{save_path}.pt"
else:
    raise ValueError(f"Unknown renewable: {self.renewable}")
#  assign data and save
sample_class._data = sample
sample_class.save(filename)

felix-e-h-p: Cheers - just updated

pvnet/data/uk_regional_datamodule.py
```diff
@@ -69,4 +50,5 @@ def _get_streamed_samples_dataset(self, start_time, end_time) -> Dataset:
 
     def _get_premade_samples_dataset(self, subdir) -> Dataset:
         split_dir = f"{self.sample_dir}/{subdir}"
-        return NumpybatchPremadeSamplesDataset(split_dir)
+        # Returns a dict of tensors
```
Sukh-P, line 53: Slight nitpick but this returns a dict of numpy arrays rather than tensors (the conversiont to tensor happens in the collate function)

Sukh-P (approved):
Looks great, thanks for adding this!

PR #317, opened 2025-02-07
Add save concurrent samples script

dfulu:
# Pull Request

## Description

- Replaces the `save_concurrent_batches.py` script with `save_concurrent_samples.py`
- Now uses `ocf-data-sampler`
- Also merged in from main

Fixes #

## How Has This Been Tested?

- Ran locally. We don't have tests for the scripts

## Checklist:


scripts/save_concurrent_samples.py
```diff
@@ -0,0 +1,187 @@
...
+class SaveFuncFactory:
+    """Factory for creating a function to save a sample to disk."""
+
+    def __init__(self, save_dir: str):
+        """Factory for creating a function to save a sample to disk."""
+        self.save_dir = save_dir
+
+    def __call__(self, sample, sample_num: int):
+        """Save a sample to disk"""
+        torch.save(sample, f"{self.save_dir}/{sample_num:08}.pt")
```
Sukh-P, line 69: Don't think this needs to be in this PR but would be good to have a think later about whether we can use the new sample classes created to use `sample.save()` again here, make it more consistent with the `save_samples.py` script as well, I think we may need to tweak the sample class slightly before doing this

dfulu: Yeh I agree, I just didn't want to do too much in this PR

PR #323, opened 2025-02-14
replaced datapipe with ocf-data-sampler

### LINKED ISSUE #313: Update backtest scripts to use ocf-data-sampler, for site
@peterdudfield you mentioned this in phase 2 in openclimatefix/ocf-data-sampler#98 
if my understaning is correct, we are trying to change ocf_datapipes to ocf-data-sampler and the changes need to made in this file https://github.com/openclimatefix/PVNet/blob/main/scripts/backtest_sites.py

any guidence or reference for getting started would be really helpful

zaryab-ali:
## Description
removed ocf_datapipe from the entire file and replaced with ocf_data_sampler, still some changes might need to be made, needs reviewing

Fixes #313

### DIFF
--- a/scripts/backtest_sites.py
+++ b/scripts/backtest_sites.py
@@ -35,30 +35,12 @@ import torch
 import xarray as xr
 from huggingface_hub import hf_hub_download
 from huggingface_hub.constants import CONFIG_NAME, PYTORCH_WEIGHTS_NAME
-from ocf_datapipes.batch import (
-    BatchKey,
-    NumpyBatch,
-    batch_to_tensor,
-    copy_batch_to_device,
-    stack_np_examples_into_batch,
-)
-from ocf_datapipes.config.load import load_yaml_configuration
-from ocf_datapipes.load.pv.pv import OpenPVFromNetCDFIterDataPipe
-from ocf_datapipes.training.common import create_t0_and_loc_datapipes
-from ocf_datapipes.training.pvnet_site import (
-    DictDatasetIterDataPipe,
-    _get_datapipes_dict,
-    construct_sliced_data_pipeline,
-    split_dataset_dict_dp,
-)
-from ocf_datapipes.utils.consts import ELEVATION_MEAN, ELEVATION_STD
+from ocf_data_sampler.torch_datasets.datasets.site import SitesDataset
 from omegaconf import DictConfig
-from torch.utils.data import DataLoader, IterDataPipe, functional_datapipe
-from torch.utils.data.datapipes.iter import IterableWrapper
+from torch.utils.data import DataLoader
 from tqdm import tqdm
 
 from pvnet.load_model import get_model_from_checkpoints
-from pvnet.utils import SiteLocationLookup
 
 # ------------------------------------------------------------------
 # USER CONFIGURED VARIABLES TO RUN THE SCRIPT
@@ -75,8 +57,8 @@ hf_token = None
 hf_model_id = None
 
 # Forecasts will be made for all available init times between these
-start_datetime = "2022-05-08 00:00"
-end_datetime = "2022-05-08 00:30"
+start_datetime = "2022-05-07 13:00"
+end_datetime = "2022-05-07 16:30"
 
 # ------------------------------------------------------------------
 # SET UP LOGGING
@@ -106,51 +88,6 @@ ALL_SITE_IDS.sort()
 
 # ------------------------------------------------------------------
 # FUNCTIONS
-
-
-@functional_datapipe("pad_forward_pv")
-class PadForwardPVIterDataPipe(IterDataPipe):
-    """
-    Pads forecast pv.
-
-    Sun position is calculated based off of pv time index
-    and for t0's close to end of pv data can have wrong shape as pv starts
-    to run out of data to slice for the forecast part.
-    """
-
-    def __init__(
-        self,
-        pv_dp: IterDataPipe,
-        forecast_duration: np.timedelta64,
-        history_duration: np.timedelta64,
-        time_resolution_minutes: np.timedelta64,
-    ):
-        """Init"""
-
-        super().__init__()
-        self.pv_dp = pv_dp
-        self.forecast_duration = forecast_duration
-        self.history_duration = history_duration
-        self.time_resolution_minutes = time_resolution_minutes
-
-        self.min_seq_length = history_duration // time_resolution_minutes
-
-    def __iter__(self):
-        """Iter"""
-
-        for xr_data in self.pv_dp:
-            t_end = (
-                xr_data.time_utc.data[0]
-                + self.history_duration
-                + self.forecast_duration
-                + self.time_resolution_minutes
-            )
-            time_idx = np.arange(xr_data.time_utc.data[0], t_end, self.time_resolution_minutes)
-
-            if len(xr_data.time_utc.data) < self.min_seq_length:
-                raise ValueError("Not enough PV data to predict")
-
-            yield xr_data.reindex(time_utc=time_idx, fill_value=-1)
 
 
 def load_model_from_hf(model_id: str, revision: str, token: str):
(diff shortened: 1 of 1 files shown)


scripts/backtest_sites.py
```diff
@@ -360,114 +166,81 @@ def __init__(self, model, ds_site: xr.Dataset):
...
-        n_valid_times = len(batch[BatchKey.pv_time_utc][0, id0 + 1 :])
-        model = self.model
+        t0 = pd.Timestamp(sample["site_init_time_utc"][0])
+        site_id = sample["site_id"][0]
 
         # Get valid times for this forecast
-        valid_times = pd.to_datetime(
-            [t0 + np.timedelta64((i + 1) * FREQ_MINS, "m") for i in range(n_valid_times)]
+        valid_times = pd.date_range(
+            start=t0 + pd.Timedelta(minutes=FREQ_MINS),
```
AUdaltsova, line 191: valid_times are supposed to be the timestamps the model makes predictions for, so they should not contain any history, i e start from t0, from where interval start and end are calculated. Judging by [this TODO](https://github.com/openclimatefix/ocf-data-sampler/blob/6dbd5fa544cdf1fc12bb40c9e4c7baeedf51b455/ocf_data_sampler/torch_datasets/datasets/site.py#L270) we might not store it in the samples right now; you might have to either make the change in the dataset itself or pull interval_start/interval_end from the data config here to get the correct times.

zaryab-ali: can you please have a look at the recent commit and let me know if the changes i made are correct or not

scripts/backtest_sites.py
```diff
@@ -183,14 +185,20 @@ def predict_batch(self, sample: dict) -> xr.Dataset:
...
         site_id = sample["site_id"][0]
 
         # Get valid times for this forecast
         valid_times = pd.date_range(
-            start=t0 + pd.Timedelta(minutes=FREQ_MINS),
-            periods=len(sample["site_target_time_utc"]),
-            freq=f"{FREQ_MINS}min",
+            start=t0 + pd.Timedelta(interval_start),
+            end=t0 + pd.Timedelta(interval_end),
+            freq=f"{time_resolution.astype(int)}min",
```
AUdaltsova, line 201: Lifting from config is a good way to do it, but the sample ModelPipe is getting already has history, so the first time in it is not t0, it's t0+interval_start. You can get t0 from last time - interval_end for example, or first time - interval_start (you can see how that happens [in this function](https://github.com/openclimatefix/ocf-data-sampler/blob/3054a5721ba2897ec76841bba97ac4c7d0e34d2c/ocf_data_sampler/select/select_time_slice.py#L8-L23))

scripts/backtest_sites.py
```diff
@@ -215,259 +153,102 @@ def get_sites_ds(config_path: str) -> xr.Dataset:
...
-        # Unpack some variables from the batch
-        id0 = batch[BatchKey.pv_t0_idx]
+        # Convert sample to tensor and move to device
+        sample_tensor = {k: torch.from_numpy(v).to(device) for k, v in sample.items()}
+
+        config = load_yaml_configuration(self.config_path)
+
+        interval_start = np.timedelta64(config.input_data.site.interval_start_minutes, "m")
+        interval_end = np.timedelta64(config.input_data.site.interval_end_minutes, "m")
+        time_resolution = np.timedelta64(config.input_data.site.time_resolution_minutes, "m")
```
AUdaltsova, line 192: Also I would probably unpack the config in the init or maybe even move the whole thing to main and pass extracted interval_end or whatever you end up using to avoid doing it again and again for every batch

scripts/backtest_sites.py
```diff
@@ -215,259 +153,102 @@ def get_sites_ds(config_path: str) -> xr.Dataset:
...
-        # We only need elevation mask for forecasted values, not history
-        elevation = elevation[:, id0 + 1 :]
+        # Get capacity for this site
+        site_capacity = float(self.ds_site.sel(site_id=site_id).capacity_kwp)
 
-        # Make mask dataset for sundown
+        # Get solar elevation and create sundown mask
+        elevation = sample["site_solar_elevation"]
         da_sundown_mask = xr.DataArray(
             data=elevation < MIN_DAY_ELEVATION,
```
AUdaltsova, line 210: There used to be a pv_system dimension which I would keep, as the intention I think is for this to work for multiple sites simultaneously, and even if we end up using just one site this functionality would still allow it

scripts/backtest_sites.py
```diff
@@ -215,259 +153,102 @@ def get_sites_ds(config_path: str) -> xr.Dataset:
...
-    # Also convert to tensor for model
-    data_pipeline = (
-        data_pipeline.batch(len(ALL_SITE_IDS))
-        .map(stack_np_examples_into_batch)
-        .map(batch_to_tensor)
-    )
-    data_pipeline = data_pipeline.set_length(num_batches)
+    # Filter for specific site IDs
+    dataset.valid_t0_and_site_ids = dataset.valid_t0_and_site_ids[
+        dataset.valid_t0_and_site_ids.site_id.isin(ALL_SITE_IDS)
```
AUdaltsova, line 248: This is a good idea but I think the selection might not work - I think at this point `dataset.valid_t0_and_site_ids` is organised as a pd.DataFrame with site_ids as column names (see [here](https://github.com/openclimatefix/ocf-data-sampler/blob/3054a5721ba2897ec76841bba97ac4c7d0e34d2c/ocf_data_sampler/torch_datasets/datasets/site.py#L145))

scripts/backtest_sites.py
```diff
@@ -215,259 +153,102 @@ def get_sites_ds(config_path: str) -> xr.Dataset:
...
-    t0_datapipe = t0_datapipe.unbatch(
-        unbatch_level=1
-    )  # might not need this part since the site datapipe is creating examples
-
-    t0_datapipe = t0_datapipe.set_length(num_t0s * len(ALL_SITE_IDS))
-    location_pipe = location_pipe.set_length(num_t0s * len(ALL_SITE_IDS))
-
-    return location_pipe, t0_datapipe
+    dataset = SitesDataset(config_path)
+    return dataset.datasets_dict["site"]
```
AUdaltsova, line 159: Instead of creating an entire dataset just for this I'd maybe just unpack the config and open the sites data directly, as in [the function that creates .datasets_dict you're using here](https://github.com/openclimatefix/ocf-data-sampler/blob/3054a5721ba2897ec76841bba97ac4c7d0e34d2c/ocf_data_sampler/load/load_dataset.py#L44-L49)

AUdaltsova (changes requested):
Great work! I've left some comments, but it's starting to look very good now.

scripts/backtest_sites.py
zaryab-ali: let me know if the following logic is ok
`last_time = pd.Timestamp(sample["site_target_time_utc"][-1])
t0 = last_time - pd.Timedelta(interval_end)

# Get valid times for this forecast
valid_times = pd.date_range(
    start=t0 + pd.Timedelta(interval_start),
    end=t0 + pd.Timedelta(interval_end),
    freq=f"{time_resolution.astype(int)}min"
)`

scripts/backtest_sites.py
zaryab-ali: `valid_t0_per_site = pd.DataFrame(index=valid_t0_times_per_site)
valid_t0_per_site["site_id"] = site_id`

i think the above code sets site_id  as column, and isin function should work properly, let me know if i am not seeing something here

zaryab-ali:
hi again @AUdaltsova sorry for bothering again, just wanted to leave a reminder that i committed the changes a few days ago so have a look whenever you have time, i understand that an ocf-data-sampler model is not available right now so if the changes look legit we can merge this in a dev branch for now and when a model does become available, we can merge it to main after some testing

AUdaltsova:
Hi @zaryab-ali, thanks for nudging me! Sorry to leave you hanging. I've taken a quick look and it looks good - I'll need to check it a bit more thoroughly before we can merge since there's no tests, and unfortunately it's a super busy time here at OCF so it might take me a bit to get back to you! Will try to do this first opportunity I get.

scripts/backtest_sites.py
```diff
@@ -215,185 +154,60 @@ def get_sites_ds(config_path: str) -> xr.Dataset:
...
         Args:
             model: PVNet site level model
-            ds_site:xarray dataset of pv site true values and capacities
+            ds_site: xarray dataset of pv site true values and capacities
         """
         self.model = model
         self.ds_site = ds_site
+        self.interval_start = interval_start
+        self.interval_end = interval_end
+        self.time_resolution = time_resolution
```
AUdaltsova, line 178: Good idea! Don't forget to update the docstring

scripts/backtest_sites.py
```diff
@@ -215,185 +154,60 @@ def get_sites_ds(config_path: str) -> xr.Dataset:
...
+        t0 = first_time - pd.Timedelta(self.interval_start)
 
-        # Get valid times for this forecast
-        valid_times = pd.to_datetime(
-            [t0 + np.timedelta64((i + 1) * FREQ_MINS, "m") for i in range(n_valid_times)]
+        # Generate valid times for inference (only t0 to t0 + interval_end)
+        valid_times = pd.date_range(
+            start=t0,
+            end=t0 + pd.Timedelta(self.interval_end),
+            freq=f"{self.time_resolution.astype(int)}min",
```
AUdaltsova, line 203: That's good! But t0 itself is not included in that range (if you look carefully at the old way this is created, it starts with t0 + freq; also, I think the way you wrote that the last timestamp will be cut off as well)

AUdaltsova:
Hi again @zaryab-ali, left a couple of comments but starting to look really good now! A couple minor changes + you have a conflict popping up. Also, please add at the top of the file's docstring something to clarify that this is an untested draft that will need further debugging and revision. 

I'll run some tests by hand as I have trouble estimating if things will work just by reading in a couple of places, feel free to wait for me to leave another comment with either changes or saying everything seems fine so you don't have to go in twice, and hopefully it's just one last round of changes before we can merge!

zakwatts:
@AUdaltsova I may be able to give a hand in testing this if you need. About to runs some experiments soon regarding this. :) Also great work @zaryab-ali, Looks like its coming along nicely and great to see you've stuck at it over the past month!

scripts/backtest_sites.py
zaryab-ali: i think the following code will do the job 


t0 = first_time - pd.Timedelta(self.interval_start)
valid_times = pd.date_range(
    start=  t0 + pd.Timedelta(self.time_resolution),
    end=  t0 + pd.Timedelta(self.interval_end),
    freq=  f"{self.time_resolution.astype(int)}min",
    inclusive="both"  # ensures the final timestamp is included
)

AUdaltsova:
Hi @zaryab-ali, just wanted to let you know that since @zakwatts has now trained a site model he'll be taking over here with testing. Thanks so much for sticking with it, great work so far! Feel free to tag me still if you have any questions/concerns

zaryab-ali:
sounds great, i hope i was of some help in the process :)

Issue #328, opened 2025-02-28
remove BatchKey in site encoders


peterdudfield:
We are trying to get ride of ocf_datapipes from this repo.
#308 

It would be great to remove it from 
https://github.com/openclimatefix/PVNet/blob/main/pvnet/models/multimodal/site_encoders/encoders.py#L7

They can be replaces with strings for now

devojyotimisra:
Hi @peterdudfield,

I would Like to solve this issue

lambaaryan011:
Hey @devojyotimisra  are you working on this or should replace me this issue

devojyotimisra:
@lambaaryan011,
The job has been done, but the PR has not been accepted yet, you can look into something else I guess...

Param1304:
Please check pull request #337

Dakshbir:
@peterdudfield 
Is this issue still open for grabs?
Can i work on this issue if the PR is not accepted yet?

peterdudfield:
> [@lambaaryan011](https://github.com/lambaaryan011), The job has been done, but the PR has not been accepted yet, you can look into something else I guess...

Could you provide a link to you PR in this issue?

Dakshbir:
@peterdudfield should i take over this issue?

Sukh-P:
This was done here https://github.com/openclimatefix/PVNet/commit/9657b20a62b8369ebc7d74ade2e0e618110204a2 so going to close this, thanks

PR #330, opened 2025-03-01
Removed "BatchKey" and fixed dependents #328

devojyotimisra:
# Pull Request

## Description

_Removed "BatchKey" and fixed dependents_

Fixes #

## How Has This Been Tested?

_Ran the command `python -m pytest tests` and all tests have passed_

- [X] Yes

_If your changes affect data processing, have you plotted any changes? i.e. have you done a quick sanity check?_

- [X] Yes

## Checklist:

- [X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/pvnet/models/multimodal/site_encoders/encoders.py
+++ b/pvnet/models/multimodal/site_encoders/encoders.py
@@ -75,11 +75,11 @@ class SimpleLearnedAggregator(AbstractSitesEncoder):
 
     def _calculate_attention(self, x):
         # added this for loop to correctly acess the enum object
-        i = 0
+        i=0
         for key in x.keys():
-            if i == 17:
-                ans = key
-            i += 1
+            if(i==17):
+                ans=key
+            i+=1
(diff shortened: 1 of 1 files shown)


devojyotimisra:
Please look into this @peterdudfield

devojyotimisra:
image
All tests have passed...

Sukh-P (changes requested):
Thanks for making a start on this, but what we meant here was actually that we have moved away from enums as keys and instead are using strings, so the strings "gsp_id" and "pv" should do for these instead

There will likely be some tests that fail as well since the test data is in the old batch key format so we will need to fix those by making some changes to the test data after it is loaded (or just changing the test data saved in the repo)

Sukh-P:
@devojyotimisra we are doing some cleaning of stale PRs, I have left some feedback on this one, if you aren't able to respond/work on this in the next week I will close the PR as you can always create a new one when/if you are able to finish this work, hope that's alright, thanks!

PR #335, opened 2025-03-11
Solar decoupling update

felix-e-h-p:
Update accounting for new decoupled solar coord keys - .pt reference files for testing updated accordingly.


pvnet/models/multimodal/multimodal.py
```diff
@@ -68,6 +68,7 @@ def __init__(
         timestep_intervals_to_plot: Optional[list[int]] = None,
         adapt_batches: Optional[bool] = False,
         forecast_minutes_ignore: Optional[int] = 0,
+        solar_position_config: Optional[dict] = None,
```
Sukh-P, line 71: So this `solar_position_config` doesn't quite make sense to me, it's currently not used anywhere in the code and I get I believe the intention of it but I think for now we can just stick with `include_sun` being the parameter which decides whether the sun data is used or not and the sun data is just used completely if it is and doesn't need any extra config, so I would remove these

tests/models/multimodal/test_multimodal.py
```diff
@@ -1,6 +1,8 @@
 from torch.optim import SGD
 import pytest
+import torch
 
+from pvnet.models.multimodal.multimodal import Model  # Import MultimodalModel
```
Sukh-P, line 5: Can remove the comment here

pyproject.toml
```diff
@@ -6,7 +6,7 @@ dynamic = ["version", "readme"]
 license={file="LICENCE"}
 
 dependencies = [
-    "ocf_data_sampler==0.1.11",
+    "ocf-data-sampler>=0.1.13",
```
Sukh-P, line 9: I think we still want to pin to a specific version of ocf-data-sampler so I would take the last version which included your changes in solar decoupling in it and pin it to that rather than >=

tests/test_end2end.py
```diff
@@ -1,9 +1,48 @@
 import lightning
+import torch
+from torch.utils.data import Dataset, DataLoader
 
 
-def test_model_trainer_fit(multimodal_model, sample_train_val_datamodule):
-    batch = next(iter(sample_train_val_datamodule.train_dataloader()))
-    y = multimodal_model(batch)
+def test_model_trainer_fit(multimodal_model, sample_batch):
```
Sukh-P, line 6: Since we already have tests including solar I don't think this test currently needs modifying, I would keep at as it was

tests/conftest.py
```diff
@@ -94,36 +94,62 @@ def sat_data():
 
 
 @pytest.fixture()
-def sample_train_val_datamodule():
-    # duplicate the sample batcnes for more training/val data
-    n_duplicates = 10
-
+def sample_train_val_datamodule(multimodal_model):
```
Sukh-P, line 97: Since this is not used I don't think we should keep these changes, I would revert it back to what it was and still have it used in the e2e test

tests/conftest.py
```diff
@@ -94,36 +96,33 @@ def sat_data():
...
+
+    # Create a dataloader that always returns sample_batch
+    dataloader = DataLoader(
+        SimpleDataset(),
+        batch_size=1,
+        collate_fn=lambda x: sample_batch
+    )
+
+    # Lightning DataModule
+    class SimpleDataModule(lightning.LightningDataModule):
```
Sukh-P, line 118: It feels like doing it this way we are essentially mocking a few parts of the training pipeline e.g. not using the actual DataModules we have and this feels against the purpose of an e2e test (where this is used)

Sukh-P: I think leaving this as it was would be fine for the scope of this PR as I don't think there would be conflicts between it and the changes you have made, but correct me if I am wrong on that

tests/models/multimodal/test_multimodal.py
```diff
@@ -45,3 +47,38 @@ def test_weighted_quantile_model_forward(multimodal_quantile_model_ignore_minute
...
+        "solar_azimuth",
+        "solar_elevation",
+    ]:
+        if key in batch_copy:
+            del batch_copy[key]
+
+    batch_size = batch_copy["gsp"].shape[0]
+    seq_len = model.forecast_len + model.history_len + 1
+    batch_copy["solar_azimuth"] = torch.rand((batch_size, seq_len))
+    batch_copy["solar_elevation"] = torch.rand((batch_size, seq_len))
```
Sukh-P, line 75: Actually looking at this again how come this is necessary, does the sample batch not already include the relevant solar keys since you added them in the fixture?

felix-e-h-p: Definitely not necessary - I suppose just to triple check RE removal and then verification. Updating now.

tests/models/multimodal/test_multimodal.py
```diff
@@ -1,6 +1,8 @@
 from torch.optim import SGD
 import pytest
+import torch
 
+from pvnet.models.multimodal.multimodal import Model
```
Sukh-P, line 5: Are these imports needed?

Sukh-P (approved):

PR #339, opened 2025-03-15
Save software versions in model card on HF for reproducibility

### LINKED ISSUE #326: Save PVNet and ocf-data-sampler/ocf_datapipes version to model card when pushing to HF
## Detailed Description
Now that we're transitioning between models that use ocf_datapipes and ones using different versions of ocf-data-sampler, it would be useful to save information on the versions of tools used to train it. I think it should go into the model card on HF under 'Software' (we currently have links to repos there, which default to ocf_datapipes regardless). I'd suggest to make the change somewhere in [here](https://github.com/openclimatefix/PVNet/blob/e9837bb60a63772e6337eeb33f97e1a6242f3057/pvnet/models/base_model.py#L309) in the model's save_pretrained method, but open to suggestions. 

Ideally, it would automatically save the versions used, but this might require changes to have we save checkpoints, so everything is stored correctly. If this is not possible, maybe it can prompt the user to input the libraries and versions on save.

chaitanya-rdY:
##*Pull Request*

Description
This PR updates the save_pretrained method in the PVNetModelHubMixin class to include version information for the tools used to train the model. This metadata will be stored in the model card on Hugging Face under the "Software" section.

Motivation & Context
Ensures reproducibility by tracking software versions.
Supports the transition between ocf_datapipes and different versions of ocf-data-sampler.

Fixes #326 

How Has This Been Tested?
Verified that save_pretrained correctly updates the model card with software version details.
✅ Sanity Check: Yes, changes have been reviewed and verified.

## **Checklist**  
- ✅ My code follows [OCF's coding style guidelines](https://github.com/openclimatefix/.github/blob/main/coding_style.md).  
- ✅ I have performed a self-review of my code.  
- ✅ I have updated the documentation accordingly.  
- ✅ I have added tests to validate this feature.  
- ✅ I have checked my code for typos and formatting errors.


chaitanya-rdY:
@AUdaltsova,this is the raised pull request for #326 issue.

pvnet/models/base_model.py
```diff
@@ -49,6 +50,9 @@ def make_clean_data_config(input_path, output_path, placeholder="PLACEHOLDER"):
     with open(input_path) as cfg:
         config = yaml.load(cfg, Loader=yaml.FullLoader)
 
+    if "general" not in config:
```
Sukh-P, line 53: Not sure what the purpose of these lines are? I think they can be removed

pvnet/models/base_model.py
```diff
@@ -320,10 +324,15 @@ def save_pretrained(
...
 
+        # Get versions of relevant packages
+        packages = ["ocf_datapipes", "ocf_data_sampler", "torch", "lightning.pytorch"]
+        package_versions = {pkg: pkg_resources.get_distribution(pkg).version for pkg in packages}
+
         card = ModelCard.from_template(
             card_data,
             template_path=card_template_path,
             wandb_links=wandb_links,
+            software_versions=package_versions,  # Add software versions to the card
```
Sukh-P, line 335: So software versions is currently not used in any of the model cards which we use so it would need to be referenced in them for this actually to work, like what is done [here](https://github.com/openclimatefix/PVNet/blob/f66e3d1cde932312bb6cf36f7cfe2118cf077df5/pvnet/models/model_cards/pv_uk_regional_model_card_template.md?plain=1#L40) where the wandb_links are referenced, this would need to be done for all three of the model cards which are in that the same folder as the file I linked

Sukh-P (changes requested):
Thanks for making a start on this, needs a few changes for this to work

pvnet/models/base_model.py
chaitanya-rdY: @Sukh-P ,If there is no "general" in the config, it creates a new one. However, during testing, it raised an error, which is why I included this.

pvnet/models/base_model.py
chaitanya-rdY: @Sukh-P , I understand what needs to be done. Can I proceed?

pvnet/models/base_model.py
Sukh-P: Hmmm I haven't seen an error around this before, please could you remove it for now and we can check the test still pass

Sukh-P (changes requested):
Just reapplying the review

Sukh-P (changes requested):
@chaitanya-rdY I can see you requested a review again but you haven't addressed the main feedback from last time yet https://github.com/openclimatefix/PVNet/pull/339/files#r2003644726 which without this being done this will not work, please address this feedback before requesting a review again, if you have questions about the feedback then please ask before writing any code

pvnet/models/base_model.py
chaitanya-rdY: @Sukh-P , Do I need to change the model card template of [PV_india_model_card](https://github.com/openclimatefix/PVNet/blob/f66e3d1cde932312bb6cf36f7cfe2118cf077df5/pvnet/models/model_cards/pv_india_model_card_template.md) to mention all software versions

chaitanya-rdY:
@Sukh-P 
We need to track the software versions used for training and reference them in the model card. Currently, the model card doesn’t include software versions, so we should modify save_pretrained to save package versions in software_versions.json, update the model card template to include {{ software_versions }} in the Software section, and update the model card generation script to load software_versions.json and insert the versions dynamically, similar to how wandb_links are handled. This will improve reproducibility and ensure that version details are automatically stored. Let me know if this approach works!

Sukh-P:
@chaitanya-rdY yes that sounds along the right lines, are you going to edit the 3 model card templates we have to include this?

pvnet/models/base_model.py
chaitanya-rdY: @Sukh-P, Ok,I will do by this week

Issue #265, opened 2024-10-22
Use pyproject.toml for dependency management and tool configuration


devsjc:
There are many files in the root of this repo dedicated to configuring tools. Since [PEP-621](https://peps.python.org/pep-0621/), a `pyproject.toml` file has become the standard for managing python projects. As such, moving what configuration files can be moved into a pyproject file, along with the dependencies and project metadata, would keep PVNet up to date with recommended packaging practices. This will also require a modification to the github actions to work with the new file in releasing.

For more details on how to consolidate configuration into pyproject, see [this blog post](https://medium.com/@devsjc/the-complete-guide-to-pyproject-toml-replace-setup-py-and-requirements-txt-553f899dc267).

chaitanya-rdY:
@devsjc,I would like to work on this issue. Could you please assign it to me?

anugrahrk:
@devsjc if this issue is still relevant, i would like to work on this.

Dakshbir:
@peterdudfield @devsjc @Sukh-P 
Can you assign this issue to me (if its still relevant)?
I believe that i will be able to fix the issue.

peterdudfield:
We now have a pyproject.toml file so I will close this issue

PR #340, opened 2025-03-15
[issue265] Clean-unify lint config files

tmi:
# Pull Request

## Description

Fixes [issue 265](https://github.com/openclimatefix/PVNet/issues/265)

Changes to pyproject and linters:
1/ Removes flake8 config -- presumably redundant after switch to ruff
2/ Moves black line length config from precommit config to pyproject
3/ Moves content from .ruff.toml into pyproject.toml
4/ Updates ruff from 0.0.286 to 0.11.0
5/ Remove unsupported `format=github` ruff config -- no idea what it was supposed to mean
6/ Moved selected options from `tool.ruff` to `tool.ruff.lint` as needed
7/ Moved `isort.cfg` to `pyproject.toml` -- however, note that isort is **not** listed in pre-commit config, and the current codebase does not seem isorted

I don't think `prettier` supports pyproject.toml so leaving that one standalone.

From what I found, `bumpversion` looks deprecated, with support for pyproject not having been added (https://github.com/c4urself/bump2version/issues/42). It should thus be replaced with eg https://github.com/callowayproject/bump-my-version -- that would however be a PR for the `openclimatefix/.github` repo

The original issue also mentions "along with the dependencies and project metadata" -- but it looks those are already in pyproject.

The original issue also mentions "This will also require a modification to the github actions to work with the new file in releasing." -- I checked the release github action and I don't think a modification is necessary (in part due to not having changed the bump2version config).

## How Has This Been Tested?

- [x] Yes: I've run `pre-commit run -a` to see no new style breach introduced, as well as making a dummy commit that broke style and then checked that `pre-commit` still catches that

## Checklist:


pyproject.toml
```diff
@@ -59,6 +59,9 @@ all_models=[
 ]
 all=["PVNet[dev,all_models]"]
 
+[tool.isort]
+profile = "black"
+
```
devsjc, line 64: If we're moving to ruff, `isort` can go altogether - ruff handles sorting by itself.

devsjc (approved):
Thanks for cleaning up the repository!

pyproject.toml
```diff
@@ -84,6 +87,58 @@ testpaths = [
     "tests",
 ]
 
+[tool.black]
```
Sukh-P, line 90: Do we need to still have black? @devsjc what do you think?

pyproject.toml
```diff
@@ -84,6 +87,58 @@ testpaths = [
...
+line-length = 100
+
 [tool.ruff]
```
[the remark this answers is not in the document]
tmi: done in another commit -- removed black, isort and flake8 altogether; added ruff to dev deps -- though not sure if really needed, since precommit installs it itself to its own venv. Do you guys want to run ruff explicitly? If not, we can leave it altogether...

Sukh-P (approved):
Great work, thanks for this!

PR #347, opened 2025-03-20
Add error distribution metrics

### LINKED ISSUE #336: Add percentile errors to validation metrics for W&B
We want to start recording percentile errors in our validation metrics which we save to W&B so that we can keep a track of large errors, e.g the 95th and 5th percentile for normalised errors (or 95th if absolute errors), may want the 98th/2nd and 90th/10th percentiles as well 

This could be in this function here https://github.com/openclimatefix/PVNet/blob/main/pvnet/models/base_model.py#L715

Sukh-P:
# Pull Request

## Description

Adding error distribution metrics which will be tracked for the validation set and saved to W&B at each epoch end, implemented a simple version which is distribution of error and absolute error over all forecast steps 

I've had a go at deciding which metrics would be useful but please feel free to suggest changes or additions to them, thanks

Fixes #336

### DIFF
--- a/pvnet/models/base_model.py
+++ b/pvnet/models/base_model.py
@@ -1,4 +1,5 @@
 """Base model for all PVNet submodels"""
+
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


pvnet/models/base_model.py
```diff
@@ -718,6 +718,27 @@ def on_validation_epoch_end(self):
...
+            validation_results_df["error"] = (
+                validation_results_df["y"] - validation_results_df["y_quantile_0.5"]
+            )
+
+            if isinstance(self.logger, pl.loggers.WandbLogger):
+                # log error distribution metrics
+                wandb.log(
+                    {
+                        "95th_percentile_error": validation_results_df["error"].quantile(0.95),
+                        "5th_percentile_error": validation_results_df["error"].quantile(0.05),
```
peterdudfield, line 730: Could it be slightly renamed so that it clear we talking about the quantile of the p50 errors. I suddenly thought this was to do with the quantile made by pvnet

peterdudfield (approved):
I've left a comment, but approved so you can adjust how you wish

pvnet/models/base_model.py
Sukh-P: Yep sure

pvnet/models/base_model.py
```diff
@@ -718,6 +718,27 @@ def on_validation_epoch_end(self):
...
             validation_results_df = pd.concat(self.validation_epoch_results)
+            validation_results_df["error"] = (
+                validation_results_df["y"] - validation_results_df["y_quantile_0.5"]
+            )
+
+            if isinstance(self.logger, pl.loggers.WandbLogger):
+                # log error distribution metrics
+                wandb.log(
+                    {
+                        "95th_percentile_median_forecast_error": validation_results_df["error"].quantile(0.95),
```
peterdudfield, line 729: do you mind order them in quantile order, then its slightly easier to read,

PR #353, opened 2025-03-25
Site encoders update

felix-e-h-p:
Removal of BatchKey from site encoders file.

Conftest updated intermediately - overwrites existing .pt reference sample - adding gsp_id and pv.

Once merged, can fully update sample_pv_batch and sample_site_batch to both utilise SiteDataModule and on the fly generated synthetic samples with no reference file or any necessity for BatchKey.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -166,8 +166,6 @@ def sample_pv_batch():
     file_path = "tests/test_data/presaved_batches/train/000000.pt"
     old_batch = torch.load(file_path)
     new_batch = {}
-    key_pv_found = False
-    key_gsp_id_found = False
 
     for key, value in old_batch.items():
         if key == BatchKey.pv:


tests/conftest.py
```diff
@@ -154,9 +155,31 @@ def sample_satellite_batch(sample_batch):
...
+
     # TODO: Once PV site inputs are available from ocf-data-sampler UK regional remove these
     # old batches. For now we use the old batches to test the site encoder models
-    return torch.load("tests/test_data/presaved_batches/train/000000.pt")
+
+    file_path = "tests/test_data/presaved_batches/train/000000.pt"
+    old_batch = torch.load(file_path)
+    new_batch = {}
+    key_pv_found = False
+    key_gsp_id_found = False
```
Sukh-P, line 170: Are these variables needed? Seems like they aren't being used for anything

Sukh-P (approved):
Thanks for updating this!

Issue #352, opened 2025-03-25
Transition from static test samples


felix-e-h-p:
We currently rely on pre-generated test samples in tests/test_data/presaved_samples_uk_regional/ for tests such as test_multimodal.py and test_end2end.py. This presented a minor issue with the recent solar data decoupling, which required (albeit minor) a direct update to the reference .pt samples.

Pretty much - static .pt test samples become outdated when any form of decoupling occurs within data-sampler and PVNet. I.e. a change in format will break numerous tests, even though fundamentally the actual data pipeline and related pre-processing stages are correct.

Specifically to update:

_Fixtures in conftest.py:_
- sample_datamodule - loads static test samples
- sample_train_val_datamodule - duplicates static test samples
- sample_batch - depends on static samples

_Tests relying on these fixtures:_
- test_unimodal_teacher.py
- test_multimodal.py
- test_end2end.py

Would just involve replacing static / reference test samples with dynamically generated fixtures that utilise the current data pipeline to create test samples on the fly - I've implemented this in a minimal sense with synthetic replicated data, just suppose now it's scaling this up to correctly replicate the testing properly.

Param1304:
@felix-e-h-p Can I work on this issue ?

felix-e-h-p:
> [@felix-e-h-p](https://github.com/felix-e-h-p) Can I work on this issue ?

Thanks Param, but this is already a work in progress. Cheers.

PR #358, opened 2025-04-07
Test sample transition

felix-e-h-p:
Full transition away from BatchKey usage and both .pt and .nc reference files for model / pipeline / end2end testing - all now on the fly with relevant types.

https://github.com/openclimatefix/PVNet/issues/352


tests/conftest.py
```diff
@@ -94,51 +89,297 @@ def sat_data():
...
+        os.makedirs(f"{tmpdirname}/train", exist_ok=True)
+        os.makedirs(f"{tmpdirname}/val", exist_ok=True)
+
+        # Generate and save synthetic samples
+        base_sample = generate_synthetic_sample()
+
+        for i in range(10):
+            # Create modified copy for each sample
+            sample = {}
+            for key, value in base_sample.items():
```
peterdudfield, line 268: is all this nesting really needed?

peterdudfield: would it better just to loop over generate_synthetic_sample()?

tests/conftest.py
```diff
@@ -94,51 +89,297 @@ def sat_data():
     return ds
 
 
+def generate_synthetic_sample():
```
peterdudfield, line 92: sohuld these functions be in ocf-data-sampler?

tests/conftest.py
```diff
@@ -94,51 +89,297 @@ def sat_data():
...
+    sample["gsp_y_osgb"] = 654321.0
+
+    # Solar position define
+    sample["solar_azimuth"] = torch.linspace(0, 180, 21)
+    sample["solar_elevation"] = torch.linspace(-10, 60, 21)
+
+    return sample
+
+
+def generate_synthetic_site_sample():
```
peterdudfield, line 162: should this be in ocf-data-sampler?

tests/conftest.py
```diff
@@ -94,51 +89,297 @@ def sat_data():
...
+    nwp_channels = np.array(['t2m', 'ssrd', 'ssr', 'sp', 'r', 'tcc', 'u10', 'v10'], dtype='<U5')
+
+    # Generate NWP data
+    nwp_init_time = pd.date_range(start=now - pd.Timedelta(hours=12), periods=1, freq="12h").repeat(50)
+    nwp_steps = pd.timedelta_range(start=pd.Timedelta(hours=0), periods=50, freq="1h")
+    nwp_data = np.random.randn(50, 8, 24, 24).astype(np.float32)
+
+    # Generate site data and solar position
+    site_data = np.random.rand(197)
+    site_lat, site_lon = 52.5, -1.5
```
peterdudfield, line 183: could you add variability here, so then you dont need to do it below?

tests/conftest.py
peterdudfield: nope as discussed

tests/conftest.py
```diff
@@ -94,51 +89,272 @@ def sat_data():
...
+        # Create train and val directories
+        os.makedirs(f"{tmpdirname}/train", exist_ok=True)
+        os.makedirs(f"{tmpdirname}/val", exist_ok=True)
+
+        # Generate and save synthetic samples
+        for i in range(10):
+            site_data = generate_synthetic_site_sample()
+
+            # Add variability to coordinates and data
+            coords_updates = {
```
peterdudfield, line 327: can you move this into `generate_synthetic_site_sample`

tests/conftest.py
```diff
@@ -94,51 +89,272 @@ def sat_data():
...
+                "site__site_id": np.int32(i % 3 + 1),
+                "site__latitude": 52.5 + i * 0.1,
+                "site__longitude": -1.5 - i * 0.05,
+                "site__capacity_kwp": 10000.0 * (1.0 + i * 0.01)
+            }
+            site_data = site_data.assign_coords(coords_updates)
+
+            # Add random noise to data variables
```
peterdudfield, line 335: why add this extra noise? is it not in `generate_synthetic_site_sample` already

tests/conftest.py
```diff
@@ -94,51 +89,272 @@ def sat_data():
...
+        yield dm
 
-        file_n = 0
 
-        for file_n, file in enumerate(
-            glob.glob("tests/test_data/presaved_samples_uk_regional/train/*.pt")
-        ):
-            sample = torch.load(file)
+@pytest.fixture()
+def sample_datamodule():
```
peterdudfield, line 282: is this the same as sample_train_val_datamodule, could then just do
```
def sample_datamodule(sample_train_val_datamodule):
    yield sample_train_val_datamodule
```

Was it the same before these changes?

peterdudfield (approved):

PR #363, opened 2025-04-11
example_configuration update dropout_timedeltas_minutes due to change in config model

AUdaltsova:
Due to [this change](https://github.com/openclimatefix/ocf-data-sampler/blob/105f21db8034be9309e8bf9d4d9c7f4af9da4ab5/ocf_data_sampler/config/model.py#L85) in ocf-data-sampler we can't have `dropout_timedeltas_minutes` be null anymore. Updating to `[]`.


zakwatts (approved):
Good spot, nice change!

PR #364, opened 2025-04-11
neater mean/std syntax in example_configuration.yaml

AUdaltsova:
Making example_configuration.yaml neater with different syntax for normalisation constants

Functionality is the same, checked the values are preserved by opening old and new versions with `load_yaml_configuration` and running `config == config_new`


zakwatts (approved):
Nice job Alex!

PR #341, opened 2025-03-16
🔄 Feature Update: Add branch parameter for HuggingFace model uploads

Dakshbir:
I've implemented the ability to specify which branch to push ML models to when uploading to HuggingFace Hub, addressing issue https://github.com/openclimatefix/PVNet/issues/325. The implementation adds a branch parameter (defaulting to "main") to the save_pretrained method in PVNetModelHubMixin and the push_to_huggingface script, enabling better tracking of model changes through git history.

## Summary of Changes

- **Added**: A `hub_branch` parameter to the `save_pretrained` method in `PVNetModelHubMixin` and the `push_to_huggingface` script.
- **Updated**: Documentation to reflect the new parameter usage.
- **Enhanced**: Git-based model versioning capabilities by allowing model pushes to a specified branch instead of defaulting to `main`.

## Why This is Needed?

Previously, all model uploads to Hugging Face Hub defaulted to the `main` branch. This led to:
- Difficulty in managing multiple versions of models within a repo.
- Lack of flexibility for teams managing experiments across branches.
- Increased risk of overwriting critical models.

By allowing users to specify a branch, we improve model version tracking and support better CI/CD workflows.

## How Has This Been Tested?

- Updated existing unit tests to include cases where the branch parameter is explicitly set.
- Ran manual tests by pushing models to different branches on Hugging Face Hub and verifying the uploaded files.
- Verified that the default behavior remains unchanged when no branch parameter is provided.

### Steps to Reproduce:
1. Call `model.save_pretrained("path/to/model", push_to_hub=True, hub_branch="new_branch")`
2. Verify that the model is pushed to the correct branch in the Hugging Face Hub.
3. Check the git commit history to confirm changes.

## Checklist:

- [ ] My code follows [OCF's coding style guidelines](https://github.com/openclimatefix/.github/blob/main/coding_style.md)
- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked my code and corrected any misspellings

Issue #325, opened 2025-02-19
Upload HF model to branch


peterdudfield:
## Detailed Description
it owuld nice to be able to upload the ML modell to HF to a specific branch, perhaps default is main.
This means we can look at the git history of the model, and see whats changed more easily. 

## Context
- currently we have 6 ml models, and the git hisotry is between the models
- the app loads from a git has, so this shouldnt have to change
- discussed with @dfulu 

## Possible Implementation
Add to branch to [here](https://github.com/openclimatefix/PVNet/blob/main/pvnet/models/base_model.py#L334) maybe. The branch name could be the model name

MAYANK12SHARMA:
Hi @peterdudfield, can i try to solve this issue?

peterdudfield:
I think @khadar1020 is going to do this. You can always work together for a solution

khadar1020:
I am working on it will update PR soon

Dakshbir:
Hi @maintainers,

I’ve submitted a PR that implements the ability to specify a branch when pushing ML models to Hugging Face Hub, addressing this issue: #325. 🚀

🔗 **PR Link:** [#341](https://github.com/openclimatefix/PVNet/pull/341)  

### **Summary of Changes**:
- Added a `hub_branch` parameter to `save_pretrained` and `push_to_huggingface` scripts.
- Updated documentation to reflect the new feature.
- Enhanced model versioning by allowing branch-specific uploads.

This should help with better tracking of model versions and improve CI/CD workflows. I'd love to get your feedback when you have time! Let me know if any improvements are needed. 😊

Thanks for your time! 🙌

Dakshbir:
Sorry, I made the PR without asking to assign the issue to me. I saw the contributions welcome label and the issue caught my eye so i resolved the issue and made the PR. 
I am extremely sorry if this caused any inconvenience.

Please review my PR and tell me any modifications

peterdudfield:
@khadar1020 are you still working on this? Or you happy for me to look at the PRs submitted by others

khadar1020:
I am happy @peterdudfield nothing to worry you can review their PR

Dakshbir:
@peterdudfield just a gentle reminder to review my PR

peterdudfield:
> @peterdudfield just a gentle reminder to review my PR
> 

I waiting for you to reply to a comment on the PR

Dakshbir:
I have seen your comment....I am fixing this....will get back to you soon🙂

[...]

PR #365, opened 2025-04-15
Assorted huggingface checkpoint upgrades

dfulu:
# Pull Request

## Description

This PR packages a few updates

- Bug: Update minimising the config for datapipes 0.2.10 when saving the model to huggingface. This was using the config variable names from an old version of data-sampler
- Enhancement: Add functionality to print huggingface commit hash when the model is updated. The printed commit hash can be directly copied from terminal and put in the config file in the app. This makes user error less likely when managing model commit numbers
- Bug: Uploading to a branch was added in #341 and designed in #325. As it turns out, huggingface does not support multiple branches and recommends 1 model 1 repo (see discussion [here](https://discuss.huggingface.co/t/different-model-card-different-branch/110029/1) for example). So I have reverted #341
- Enhancement: Slight changes to how we use typer in the upload to huggingface script. Typer prints way too much when the function fails, and it then really hard to debug. This reduces what typer prints significantly.

## How Has This Been Tested?

I've ran these recently for the last round of PVNet UK model uploads

## Checklist:


Sukh-P (approved):
Nice! Thanks for making these changes

scripts/checkpoint_to_huggingface.py
```diff
@@ -24,7 +26,7 @@ def push_to_huggingface(
     wandb_ids: list[str] = [],
     local_path: str = None,
     push_to_hub: bool = True,
-    revision: str = "main",
+    revision: str | None = None,
```
Sukh-P, line 29: Previously there were some issues with the | sytanx being supported by `typer`, I am guessing this works now?

PR #366, opened 2025-04-17
Config validation

felix-e-h-p:
Specific validation function and related helper functions within new .py file for multimodal config. 
Relevant tests added, alongside additional conftest fixture.


felix-e-h-p:
Likewise with data-sampler, considering validation in regards to defining all expected keys and warn about any unexpected parameters - perhaps to consider unused settings also?

Maybe also an additional check for specific param value constraints such as ensuring time values are positive or dropout rates are within valid ranges etc.

Hope conftest fixture all OK - not cleanest, but outside of using a defined reference yaml it seems sufficient as is.

pvnet/models/multimodal/config_validation.py
```diff
@@ -0,0 +1,535 @@
...
+                   numeric types. Note: Non-critical type mismatches for some time
+                   parameters might only log warnings.
+        ValueError: If certain values are invalid, such as 'output_quantiles' being
+                    an empty list, numeric encoder parameters (channels, features,
+                    image sizes, etc.) being non-positive, or if keys within NWP
+                    time parameter dictionaries ('nwp_history_minutes',
+                    'nwp_forecast_minutes') do not exactly match the sources defined
+                    in 'nwp_encoders_dict'.
+    """
+    _check_key(cfg, "_target_", required=True, expected_type=str)
```
peterdudfield, line 452: I think you can skip `_target_`

peterdudfield (commented):
I think this might need a bit more a re-write. 

I was expecting a sample to be used in this configuration. 
```validate(numpy_batch:dict, multimodal_config:dict)```

the numpy_batch is a group of sample after `to_numpy` is applied to them. There might be some examples in conftest. 

Then in the validation, we should check that the shape of data is correct, for
-  satellite
- various nwp
- gsp
- site
- other features?

the shape should be checked using the variables in `multimodal_config`.

**Comment by felix-e-h-p** [comment]
> I think this might need a bit more a re-write.
> 
> I was expecting a sample to be used in this configuration. `validate(numpy_batch:dict, multimodal_config:dict)`
> 
> the numpy_batch is a group of sample after `to_numpy` is applied to them. There might be some examples in conftest.
> 
> Then in the validation, we should check that the shape of data is correct, for
> 
> * satellite
> * various nwp
> * gsp
> * site
> * other features?
> 
> the shape should be checked using the variables in `multimodal_config`.

New function defined as validate - defined to accept numpy_batch: NumpyBatch.
Suppose core purpose of the function is to validate shapes of data arrays within NumpyBatch - checks the shapes for:
satellite_actual (if sat_encoder is configured), NWP sources (nwp_encoders_dict if configured), pv (if pv_encoder is configured - uses num_sites here), gsp (if include_gsp_yield_history is enabled) and solar features (solar_azimuth, solar_elevation).
Consistently uses params from multimodal_config dictionary.

File: pvnet/models/multimodal/config_validation.py [hunk 2]
```diff
@@ -0,0 +1,846 @@
...
+                section_key="nwp_encoders_dict",
+                context=f"nwp_encoders_dict[{source}]",
+                source_key=source,
+            )
+
+
+def validate(
+    numpy_batch: NumpyBatch,
+    multimodal_config: dict,
+    default_interval_minutes: int = 5,
```
peterdudfield, line 625: could `default_interval_minutes` in `multimodal_config`

pvnet/models/multimodal/config_validation.py
```diff
@@ -0,0 +1,846 @@
...
+    # Satellite validation
+    if "sat_encoder" in cfg and cfg.get("sat_encoder"):
+        key = "satellite_actual"
+        if key not in numpy_batch:
+            raise KeyError(f"Batch missing '{key}' data")
+        data = numpy_batch[key]
+        if not isinstance(data, np.ndarray):
+            raise TypeError(f"'{key}' data must be np.ndarray")
+
+        interval = _get_modality_interval(cfg, key, default_interval_minutes)
```
peterdudfield, line 661: different modalities will have different interval minutes e.g
satellite = 5 mins
nwp = 60 mins
GSP = 30 mins
PV = 5 mins

pvnet/models/multimodal/config_validation.py
```diff
@@ -0,0 +1,846 @@
...
+        c = sat_cfg["in_channels"]
+        hist_steps, _ = _get_time_steps(cfg["sat_history_minutes"], 0, interval)
+        expected_shape = (hist_steps, c, h, w)
+
+        if data.ndim != 5 or data.shape[1:] != expected_shape:
+            raise ValueError(
+                f"'{key}' shape error using interval {interval}. "
+                "Expected B x {expected_shape}, Got {data.shape}"
+            )
+        batch_size = _check_batch_size_consistency(batch_size, data, key)
```
peterdudfield, line 674: nice

tests/conftest.py
```diff
@@ -89,6 +85,84 @@ def sat_data():
...
+        },
+
+        "satellite_time_resolution_minutes": 5,
+        "gsp_time_resolution_minutes": 30,
+        "sun_time_resolution_minutes": 30,
+    }
+    return cfg
+
+
+def _convert_to_numpy_recursive(item):
```
peterdudfield, line 143: do we have this already, 

I think so, could be `from ocf_data_sampler.numpy_sample.collate import stack_np_samples_into_batch`

tests/conftest.py
```diff
@@ -100,43 +174,28 @@ def generate_synthetic_sample():
...
-            "nwp_step": torch.arange(12, dtype=torch.float32),
-            "nwp_target_time_utc": torch.tensor(
-                [(now + pd.Timedelta(hours=i)).timestamp() for i in range(12)]
-            ),
+            "nwp_target_time_utc": torch.tensor([(now + pd.Timedelta(hours=i)).timestamp() for i in range(11)]),
         },
     }
 
     # Satellite define
+    sample["satellite_pred"] = {
```
peterdudfield, line 192: remove, see above comment

tests/models/multimodal/test_config_validation.py
```diff
@@ -0,0 +1,439 @@
...
+        if data.ndim == 0:
+            raise ValueError(f"Data for key '{key}' is a 0-dim array, cannot get batch size.")
+        return data.shape[0]
+    else:
+        raise TypeError(
+            f"Data for key '{key}' ({type(data)}) is not a numpy array or NWP dict."
+        )
+
+
+def test_validate_batch_error_batch_size_inconsistency(
```
peterdudfield, line 325: change this to `test_validate_batch_size`, 
- in validate function put in batch size
- check batch size when checking shapre
- this test can check errors are raised

pvnet/models/multimodal/config_validation.py
peterdudfield: batch size could be optional

tests/conftest.py
```diff
@@ -89,6 +88,148 @@ def sat_data():
...
+
+        "satellite_time_resolution_minutes": 5,
+        "gsp_time_resolution_minutes": 30,
+        "sun_time_resolution_minutes": 30,
+    }
+    return cfg
+
+
+@pytest.fixture
+def valid_input_data_config():
```
peterdudfield, line 147: could you just use the default `Configuration` instead

tests/models/multimodal/test_config_validation.py
```diff
@@ -0,0 +1,529 @@
...
+from pvnet.models.multimodal.config_validation import validate
+
+NumpyBatch = dict
+logger = logging.getLogger(__name__)
+
+
+def test_validate_valid_inputs(
+    valid_config_dict,
+    sample_numpy_batch,
+    valid_input_data_config
```
peterdudfield, line 18: use `Configuration` default here instead

pvnet/models/multimodal/config_validation.py
```diff
@@ -0,0 +1,1007 @@
...
+        multimodal_config: The multimodal configuration dictionary (for model settings).
+        input_data_config: The input data configuration dictionary (for data source settings).
+        expected_batch_size: Optional expected batch size to validate against.
+
+    Raises:
+        KeyError, TypeError, ValueError: If config is invalid or batch shapes mismatch config.
+    """
+
+    try:
+        _validate_static_config(multimodal_config)
```
peterdudfield, line 644: could intitla the model here. As long as it doesnt take to long / use too much memory. There might be less code to maintain then

pvnet/models/multimodal/config_validation.py
```diff
@@ -0,0 +1,1007 @@
...
+        c = sat_cfg["in_channels"]
+        hist_mins = cfg.get("sat_history_minutes", cfg["history_minutes"])
+        hist_steps, _ = _get_time_steps(hist_mins, 0, interval)
+        expected_shape_no_batch = (hist_steps, c, h, w)
+
+        if data.ndim != 5:
+             raise ValueError(
+                f"'{key}' dimension error. Expected 5 dims (B, T, C, H, W), Got {data.ndim}"
+            )
+        if data.shape[1:] != expected_shape_no_batch:
```
peterdudfield, line 692: could use
`expected_shape_no_batch = (batch_size, hist_steps, c, h, w)`
and
`data.shape != expected_shape_no_batch`

pvnet/models/multimodal/config_validation.py
```diff
@@ -0,0 +1,1007 @@
...
+            raise ValueError(
+                f"'{key}' shape error using interval {interval}. "
+                f"Expected B x {expected_shape_no_batch}, Got B x {data.shape[1:]}"
+            )
+        inferred_batch_size = _check_batch_size_consistency(
+            inferred_batch_size, data, key
+        )
```
peterdudfield, line 699: remove this is do above

pvnet/models/multimodal/config_validation.py
```diff
@@ -0,0 +1,1007 @@
...
+            expected_shape_no_batch = (hist_steps + forecast_steps, c, h, w)
+
+            if source_data_array.ndim != 5:
+                msg = (
+                    f"NWP source '{source}' dimension error. Expected 5 dims "
+                    f"(B, T, C, H, W), Got {source_data_array.ndim}"
+                )
+                raise ValueError(msg)
+
+            if source_data_array.shape[1:] != expected_shape_no_batch:
```
peterdudfield, line 809: could do same as above

pvnet/models/multimodal/config_validation.py
```diff
@@ -0,0 +1,1007 @@
...
+import logging
+from typing import Any, Optional, Type
+
+import numpy as np
+from ocf_data_sampler.torch_datasets.sample.base import NumpyBatch
+
+logger = logging.getLogger(__name__)
+
+
+def _check_key(
```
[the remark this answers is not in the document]
peterdudfield: validate/checks.py
validate/sample.py

tests/models/multimodal/test_config_validation.py
felix-e-h-p: Updated conftest so all tests automatically use updated fixture - derives value from the default Configuration

felix-e-h-p:
Currently have kept the separate _validate_static_config, primarily for efficiency and perhaps more targeted error messages for config related issues. Good for a first pass right to catch potential structural / type errors before attempting batch validation. Suppose can be directly replaced with model instantiation with the current validation flow here - but would perhaps like to include that as potentially another validation script if OK? Maybe another PR for that.

tests/conftest.py
```diff
@@ -100,43 +174,28 @@ def generate_synthetic_sample():
...
-            "nwp_init_time_utc": torch.tensor(
-                [(now - pd.Timedelta(hours=i)).timestamp() for i in range(11)]
-            ),
+            "nwp_init_time_utc": torch.tensor([(now - pd.Timedelta(hours=i)).timestamp() for i in range(11)]),
             "nwp_step": torch.arange(11, dtype=torch.float32),
-            "nwp_target_time_utc": torch.tensor(
-                [(now + pd.Timedelta(hours=i)).timestamp() for i in range(11)]
-            ),
-        },
-        "sat_pred": {
```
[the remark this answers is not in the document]
peterdudfield: not done

tests/conftest.py
```diff
@@ -89,6 +98,109 @@ def sat_data():
...
+    return numpy_dict
+
+
+@pytest.fixture()
+def sample_numpy_batch() -> NumpyBatch:
+    batch_size = 4
+    sample_list = []
+    for i in range(batch_size):
+        pytorch_sample = generate_synthetic_sample()
+        numpy_sample = convert_pytorch_dict_to_numpy(pytorch_sample)
```
peterdudfield, line 196: can you try without `convert_pytorch_dict_to_numpy`

pvnet/models/multimodal/validation/config_validation.py
```diff
@@ -0,0 +1,189 @@
...
+    except (KeyError, TypeError, ValueError) as e:
+        logger.error(f"Static configuration validation failed: {e}")
+        raise
+
+    logger.info(
+        f"Validating batch data shapes against configuration. Expected batch size: "
+        f"{expected_batch_size if expected_batch_size is not None else 'Not Provided'}."
+    )
+    cfg = multimodal_config
+    inferred_batch_size: int | None = None
```
peterdudfield, line 42: remove

pvnet/models/multimodal/validation/config_validation.py
```diff
@@ -0,0 +1,189 @@
...
+        key = "satellite_actual"
+        logger.debug(f"Validating modality: {key}")
+        data = _check_batch_data(numpy_batch, key, np.ndarray, config_key_sat)
+        interval = _get_modality_interval(input_data_config, "satellite")
+        sat_cfg = _get_encoder_config(cfg, config_key_sat, config_key_sat)
+        h = w = sat_cfg["image_size_pixels"]
+        c = sat_cfg["in_channels"]
+        hist_mins = cfg.get("sat_history_minutes", cfg["history_minutes"])
+        hist_steps, _ = _get_time_steps(hist_mins, 0, interval)
+        expected_shape_no_batch = (hist_steps, c, h, w)
```
peterdudfield, line 56: `expected_shape = (expected_batch_size, hist_steps, c, h, w)`

pvnet/models/multimodal/validation/config_batch_validation.py
```diff
@@ -0,0 +1,208 @@
...
+            f"Could not find 'time_resolution_minutes' for {error_context} "
+            f"in input_data_config"
+        )
+    except ValueError as e:
+        raise ValueError(f"Invalid time_resolution_minutes for {error_context}: {e}")
+
+
+def _validate_array_shape(
+    data: np.ndarray,
+    expected_ndim: int,
```
peterdudfield, line 95: could remove, and get from `expected_shape_no_batch`

pvnet/models/multimodal/validation/config_validation.py
```diff
@@ -0,0 +1,189 @@
...
+        )
+        expected_shape_no_batch = (hist_steps + forecast_steps,)
+        for key_sun in ["solar_azimuth", "solar_elevation"]:
+            data_sun = _check_batch_data(numpy_batch, key_sun, np.ndarray, config_key_sun)
+            inferred_batch_size = _validate_array_shape(
+                data_sun, 2, expected_shape_no_batch, key_sun, sun_interval, inferred_batch_size
+            )
+        logger.debug(f"'sun' validation passed with interval {sun_interval}.")
+
+    # Final Batch Size Check
```
peterdudfield, line 164: can remove

pvnet/models/multimodal/validation/config_validation.py
```diff
@@ -0,0 +1,189 @@
...
+
+def validate(
+    numpy_batch: NumpyBatch,
+    multimodal_config: dict,
+    input_data_config: dict,
+    expected_batch_size: Optional[int] = None,
+) -> None:
+    """Validate a batch of numpy data against the multimodal configuration using helpers."""
+    try:
+        _validate_static_config(multimodal_config)
```
peterdudfield, line 32: like this split

pvnet/models/multimodal/validation/config_validation.py
peterdudfield: Also remove `5`, see comment above

tests/conftest.py
```diff
@@ -89,6 +98,109 @@ def sat_data():
     return ds
 
 
+@pytest.fixture
+def valid_config_dict(model_minutes_kwargs):
+    cfg = {
```
peterdudfield, line 103: could change to yaml, to make it neater

pvnet/models/multimodal/validation/config_batch_validation.py
```diff
@@ -0,0 +1,194 @@
...
+            f"'{batch_key}' data must be {expected_type_name}, "
+            f"found {type(data).__name__}"
+        )
+    return data
+
+
+def get_modality_interval(
+    input_data_config: dict,
+    primary_modality_key: str,
+    secondary_modality_key: Optional[str] = None,
```
AUdaltsova, line 37:
```suggestion
    secondary_modality_key: str | None,
```

I think we generally try to stick to the newer union syntax elswere in the codebase (or at least in dsampler)

pvnet/models/multimodal/validation/config_batch_validation.py
```diff
@@ -0,0 +1,194 @@
...
+        raise KeyError(
+            f"Batch missing required '{batch_key}' data "
+            f"(required by config key '{context_config_key}')."
+        )
+    data: Any = numpy_batch[batch_key]
+    if not isinstance(data, expected_type):
+        expected_type_name = getattr(expected_type, '__name__', str(expected_type))
+        raise TypeError(
+            f"'{batch_key}' data must be {expected_type_name}, "
+            f"found {type(data).__name__}"
```
AUdaltsova, line 29: I'm not a type expert so probably just something I don't know about, but why are we protecting against type not having `__name__` for expected type but not for found type?

pvnet/models/multimodal/validation/config_batch_validation.py
```diff
@@ -0,0 +1,194 @@
...
+    return data
+
+
+def get_modality_interval(
+    input_data_config: dict,
+    primary_modality_key: str,
+    secondary_modality_key: Optional[str] = None,
+    is_nwp_source: bool = False,
+) -> int:
+    """Get and validate time resolution interval from input_data_config."""
```
AUdaltsova, line 40: I think argument docstrings would be good to have here and in the above function as well. I'm nitpicking, but the argument names are not super obvious to me coming into this PR fresh, and given that these functions are likely the first thing a person would look into for debugging if the errors are not enough, I think it would be good to be very clear what they're checking for etc.

pvnet/models/multimodal/validation/config_validation.py
```diff
@@ -0,0 +1,189 @@
...
+    Validate a batch of numpy data against the multimodal configuration using helpers.
+
+    Validates shapes and types. If expected_batch_size is provided, validates
+    batch dimension against it. Does not infer or enforce cross-modality batch size
+    consistency if expected_batch_size is None.
+    """
+    if not isinstance(expected_batch_size, int) or expected_batch_size <= 0:
+        raise ValueError(
+            f"expected_batch_size must be a positive integer, got {expected_batch_size}"
+        )
```
AUdaltsova, line 39: The docstring sounds like batch size is optional, but the function requires it and this check will fail if it's None or something, is it supposed to be optional?

pvnet/models/multimodal/validation/config_validation.py
```diff
@@ -0,0 +1,189 @@
...
+    try:
+        validate_static_config(multimodal_config)
+    except (KeyError, TypeError, ValueError) as e:
+        logger.error(f"Static configuration validation failed: {e}")
+        raise
```
AUdaltsova, line 44: This is probably a me-problem, but I'm not sure what static configuration is supposed to mean in this context? From reading the function itself, it's trying to check for any errors in the model config, right?

pvnet/models/multimodal/validation/config_static_validation.py
```diff
@@ -0,0 +1,514 @@
...
+        cfg,
+        "pv_history_minutes",
+        owner_key="pv_encoder",
+        required_if_owner_present=True,
+    )
+    if pv_section:
+        _check_attention_encoder_params(
+            cfg, section_key="pv_encoder", context="pv_encoder"
+        )
+
```
AUdaltsova, line 501: I don't think there is a check for the wind encoder, is that by design?

pvnet/models/multimodal/validation/config_static_validation.py
```diff
@@ -0,0 +1,514 @@
+"""Validation functoions for Multimodal configuration."""
```
AUdaltsova, line 1:
```suggestion
"""Validation functions for Multimodal configuration."""
```

pvnet/models/multimodal/validation/config_batch_validation.py
felix-e-h-p: expected_type can be list[int]) right, that might lack a __name__ attribute. Found type always has a __name__ attribute. At leat I think anyway. Will double check, but may be the most 'robust' as it is.

pvnet/models/multimodal/validation/config_static_validation.py
```diff
@@ -0,0 +1,514 @@
...
+    section_content: Any = cfg.get(section_name)
+
+    # Ensure section follows dict structure
+    if not isinstance(section_content, dict):
+        raise TypeError(
+            f"Config section '{section_name}' must be a dictionary, "
+            f"found {type(section_content).__name__}."
+        )
+
+    if not section_content and not required:
```
AUdaltsova, line 101: I don't think empty content will ever get to this check? It should return `False` for `isinstance(section_content, dict)` in the previous one?

pvnet/models/multimodal/validation/config_validation.py
felix-e-h-p: It is! Like whether required keys are present in the configuration etc

pvnet/models/multimodal/validation/config_static_validation.py
```diff
@@ -0,0 +1,514 @@
...
+        logger.warning(
+            f"Optional config section '{section_name}' is present but empty."
+        )
+        return section_content
+
+    if check_sub_items_target:
+        if not section_content and required:
+             # If required and needs sub-items, it cannot be empty
```
AUdaltsova, line 109: I might be forgetting some edge-case here, but it feels strange that it's only checked if sub items are required to have targets? As far as I can see, the tow required sections are output network and optimizer, and I think both of them are not allowed to be empty

pvnet/models/multimodal/validation/config_static_validation.py
```diff
@@ -0,0 +1,514 @@
...
+            if not isinstance(sub_config, dict):
+                raise TypeError(
+                    f"Config for source '{source}' in '{section_name}' "
+                    f"must be a dictionary, found {type(sub_config).__name__}."
+                )
+            if "_target_" not in sub_config:
+                raise KeyError(
+                    f"Source '{source}' in section '{section_name}' "
+                    f"missing required sub-key: '_target_'"
+                )
```
AUdaltsova, line 124: Possibly controversial, but I think sub-items are only used for NWPs, and it feels strange to me that they are treated as a special thing when actually they are just regular encoder sections wrapped into a dict. It might be more consistent to treat them as their own dict sections when we iterate through the sources on L508-514 instead?

pvnet/models/multimodal/validation/config_static_validation.py
```diff
@@ -0,0 +1,514 @@
...
+    elif check_target:
+         # Only check target if section is not empty or if it's required (implicitly not empty)
+         if section_content or required:
```
AUdaltsova, line 127: I think it should never reach this point if it's empty at all. If it's not required and empty there's a check above that should've exited already, if it's empty and required I think it should be checked for explicitly and earlier than here

pvnet/models/multimodal/validation/config_static_validation.py
AUdaltsova: Sorry we use site instead of wind now, my bad! I think site is missing though

pvnet/models/multimodal/validation/config_static_validation.py
```diff
@@ -0,0 +1,514 @@
...
+                logger.warning(
+                    f"{message} (Action: May use default value if applicable)."
+                )
+        else:
+            _check_key(
+                cfg,
+                param_name,
+                required=False,
+                expected_type=int,
+                warn_on_type_mismatch=True,
```
AUdaltsova, line 172: Given all time parameters are required, I think this should fail on type mismatch if it gets here (so, both parent and the time parameter are present)

pvnet/models/multimodal/validation/config_static_validation.py
```diff
@@ -0,0 +1,514 @@
...
+    # Check time param values are integers
+    _check_key(
+        cfg, "nwp_history_minutes", required=True, expected_type=dict, context=context
+    )
+    _check_key(
+        cfg, "nwp_forecast_minutes", required=True, expected_type=dict, context=context
+    )
+    _check_key(
+        cfg, "nwp_interval_minutes", required=True, expected_type=dict, context=context
+    )
```
AUdaltsova, line 205: I think this would let through a dict with a bunch of bad values, which is not what we want

pvnet/models/multimodal/validation/config_static_validation.py
```diff
@@ -0,0 +1,514 @@
...
+
+    nwp_hist_times: dict[str, Any] = cfg["nwp_history_minutes"]
+    nwp_forecast_times: dict[str, Any] = cfg["nwp_forecast_minutes"]
```
AUdaltsova, line 208:
```suggestion
    nwp_hist_times: dict[str, int] = cfg["nwp_history_minutes"]
    nwp_forecast_times: dict[str, int] = cfg["nwp_forecast_minutes"]
```
I thought they were only allowed to be ints? Correct me if I'm wrong!

pvnet/models/multimodal/validation/config_static_validation.py
AUdaltsova: Oh I think there's a check further down that deals with that, my bad!

pvnet/models/multimodal/validation/config_static_validation.py
```diff
@@ -0,0 +1,514 @@
...
+        raise ValueError(
+            f"Keys in 'nwp_forecast_minutes' {forecast_keys} do not match sources "
+            f"in 'nwp_encoders_dict' {encoder_keys}. "
+            f"Missing: {missing_in_forecast}, Extra: {extra_in_forecast}"
+        )
+
+    _check_key(
+        cfg, "nwp_interval_minutes", required=True, expected_type=dict, context=context
+    )
+    nwp_intervals: dict[str, Any] = cfg["nwp_interval_minutes"]
```
AUdaltsova, line 235:
```suggestion
    nwp_intervals: dict[str, int] = cfg["nwp_interval_minutes"]
```

I think this is required to be an int

pvnet/models/multimodal/validation/config_static_validation.py
```diff
@@ -0,0 +1,514 @@
...
+) -> None:
+    """Check if all values in a dictionary are integers (logs warning on mismatch)."""
+    expected_type = int
+    for source, value in data.items():
+        if not isinstance(value, expected_type):
+            message = (
+                f"'{dict_name}' for source '{source}' expected {expected_type.__name__}, "
+                f"found {type(value).__name__}."
+            )
+            logger.warning(message)
```
AUdaltsova, line 345: I think this is only used for nwp parameters and should fail hard because we require all of them to be ints

pvnet/models/multimodal/validation/config_static_validation.py
```diff
@@ -0,0 +1,514 @@
...
+    _check_key(
+        cfg,
+        "output_quantiles",
+        required=True,
+        expected_type=(list, tuple),
+        context=context,
+    )
+    quantiles = cfg["output_quantiles"]
+    if not quantiles:
+        raise ValueError(f"{context}: 'output_quantiles' list cannot be empty.")
```
AUdaltsova, line 275: I am not sure if we allow quantiles to be null, I was under the impression this was a valid setting if you want to default to MAE instead of quantile loss? But I don't think we ever use that to be fair

pvnet/models/multimodal/validation/config_static_validation.py
```diff
@@ -0,0 +1,514 @@
...
+            f"{context}: Final resolved encoder config is not a dictionary."
+        )
+
+    return encoder_config
+
+
+def _check_positive_int_params_in_dict(
+    config_dict: dict[str, Any], param_names: list[str], context: str
+) -> None:
+    """Check if specified keys in a dict exist, are integers, and are positive.
```
AUdaltsova, line 405:
```suggestion
    """Check if specified keys in a dict exist, their values are integers, and are positive.
```

pvnet/models/multimodal/validation/config_static_validation.py
felix-e-h-p: Ahh yeah - picked up in _check_dict_values_are_int right. This will be altered.

pvnet/models/multimodal/validation/config_batch_validation.py
```diff
@@ -0,0 +1,269 @@
...
+            expected to hold 'time_resolution_minutes'.
+        primary_modality_key: The main key for the modality (e.g., "satellite",
+            "nwp", "sun") in `input_data_config`.
+        secondary_modality_key: An optional secondary key. If `is_nwp_source`
+            is True, this specifies the NWP source (e.g., "ecmwf"). If
+            `primary_modality_key` is "sun", this can specify a fallback
+            modality (e.g., "gsp") for interval lookup.
+        is_nwp_source: Boolean flag. If True, indicates that the primary key
+            is "nwp" and the `secondary_modality_key` refers to a specific
+            NWP data source within the "nwp" section of the config.
```
AUdaltsova, line 69: I'm nitpicking but I think this flag is unnecessary. There are no intervals that are not source-specific, so basically any time that the primary key is NWP we'd expect a secondary key for the actual source I think

pvnet/models/multimodal/validation/config_batch_validation.py
```diff
@@ -0,0 +1,269 @@
...
+            the expected shape.
+    """
+    if not isinstance(data, np.ndarray):
+        raise TypeError(f"'{data_key}' data must be NumPy array, found {type(data).__name__}")
+    if data.ndim == 0:
+        raise ValueError(f"'{data_key}' array is scalar.")
+
+    actual_batch_size = data.shape[0]
+    if actual_batch_size <= 0:
+        raise ValueError(f"'{data_key}' batch dimension has size <= 0: {actual_batch_size}")
```
AUdaltsova, line 167: Purely out of curiosity, is numpy capable of returning a negative size?

pvnet/models/multimodal/validation/config_batch_validation.py
```diff
@@ -0,0 +1,269 @@
...
+        allowed_ndims_str = " or ".join(map(str, sorted(list(allowed_ndims))))
+        raise ValueError(
+            f"'{data_key}' dimension error. Expected {allowed_ndims_str} dims, Got {actual_ndim}"
+        )
+
+    if actual_ndim == expected_ndim_base:
+        if data.shape != expected_shape_with_batch:
+             raise ValueError(
+                 f"'{data_key}' shape error using interval {interval}. "
+                 f"Expected {expected_shape_with_batch}, Got {data.shape}"
```
AUdaltsova, line 193: This is the part I mean re: saying interval is confusing. But to be honest, I don't know if stating the time resolution for context is that helpful. It's only one parameter in the time dimension (it won't help you much if you don't know the actual time interval for the data), and tells you nothing about the expected spatial crop. Is there a reason the dimensions are checked all together instead of one by one? I'd expect them to be matched 1:1 to be able to give ,ore informative errors about what specifically is wrong, but maybe there's a reason this doesn't happen that I'm not seeing?

pvnet/models/multimodal/validation/config_validation.py
```diff
@@ -0,0 +1,192 @@
...
+        logger.debug(f"Validating modality: {key}")
+        data = check_batch_data(numpy_batch, key, np.ndarray, config_key_sat)
+        interval = get_modality_interval(
+            input_data_config,
+            "satellite",
+            secondary_modality_key=None)
+        sat_cfg = get_encoder_config(cfg, config_key_sat, config_key_sat)
+        h = w = sat_cfg["image_size_pixels"]
+        c = sat_cfg["in_channels"]
+        hist_mins = cfg.get("sat_history_minutes", cfg["history_minutes"])
```
AUdaltsova, line 65: I think this shouldn't default to history_minutes, we don't allow it anymore

pvnet/models/multimodal/validation/config_validation.py
```diff
@@ -0,0 +1,192 @@
...
+        hist_steps, _ = get_time_steps(hist_mins, 0, interval)
+        expected_shape_no_batch = (hist_steps, c, h, w)
+        full_expected_shape = (expected_batch_size,) + expected_shape_no_batch
+        validate_array_shape(
+            data, full_expected_shape, key, interval
+        )
+        logger.debug(f"'{key}' shape validation passed (interval {interval}).")
```
AUdaltsova, line 72: Same re: interval as only context, but maybe I'm not seeing something

pvnet/models/multimodal/validation/config_validation.py
```diff
@@ -0,0 +1,192 @@
...
+            logger.warning(
+                f"NWP batch has extra sources not in config: {batch_sources - config_sources}"
+            )
+        nwp_hist_mins = cfg["nwp_history_minutes"]
+        nwp_forecast_mins = cfg["nwp_forecast_minutes"]
+        for source in config_sources:
+            source_key_str = f"{key}[{source}]"
+            logger.debug(f"Validating NWP source: {source}")
+            source_data_array = validate_nwp_source_structure(nwp_batch_data, source)
+            interval = get_modality_interval(input_data_config, "nwp", source, is_nwp_source=True)
```
AUdaltsova, line 97: We actually [specify time resolutions for nwp in the model config](https://github.com/openclimatefix/PVNet/blob/a9a1a3a20c5521c13ececa77fe94e23e88dd3bb3/configs.example/model/multimodal.yaml#L93C1-L93C21) so I think they should be lifted from their and default to 60 to replicate the model behaviour so we can validate the parameter that is being used

pvnet/models/multimodal/validation/config_validation.py
```diff
@@ -0,0 +1,192 @@
...
+             raise KeyError(
+                 "Neither 'site' nor 'pv' section found as a dictionary in "
+                 "input_data_config for PV validation"
+             )
+        logger.debug(f"Validating modality: {key} using input config '{modality_data_cfg_key}'")
+        data = check_batch_data(numpy_batch, key, np.ndarray, config_key_pv)
+        interval = get_modality_interval(input_data_config, modality_data_cfg_key)
+        pv_cfg = get_encoder_config(cfg, config_key_pv, config_key_pv)
+        num_sites = pv_cfg["num_sites"]
+        hist_mins = cfg.get("pv_history_minutes", cfg["history_minutes"])
```
AUdaltsova, line 133: We don't allow defaulting here either

pvnet/models/multimodal/validation/config_validation.py
```diff
@@ -0,0 +1,192 @@
...
+        fallback_key = next(
+            (k for k in possible_fallback_keys if k in input_data_config and
+                isinstance(input_data_config.get(k), dict)),
+            None
+        )
+        if fallback_key is None:
+            raise KeyError(
+                 "Cannot determine interval for sun: No suitable fallback modality "
+                 "('gsp', 'site', or 'pv') found as dictionary in input_data_config."
+             )
```
AUdaltsova, line 178: That seems a little strange to me. I agree that we wouldn't expect to find include_sun=True and not get any of the pv sources in the input data, but logically I would expect this check to go: include sun? yes. sun config exists? yes -> proceed with sun config. no? do the fallbacks exist? yes -> proceed with fallbacks. no -> fail
Generally speaking, the only time none of the fallbacks are present would be an old wind data config I think, so basically you'd just have no target data at all

tests/test_valid_config_dict.yaml
```diff
@@ -0,0 +1,71 @@
...
+  lr: 0.0001
+  weight_decay: 0.01
+  amsgrad: True
+  patience: 5
+  factor: 0.1
+  threshold: 0.002
+
+satellite_time_resolution_minutes: 5
+gsp_time_resolution_minutes: 30
+sun_time_resolution_minutes: 30
```
AUdaltsova, line 71: Sorry I must've missed an update where this was added and having trouble locating where these parameters are used now, would you mind pointing me?

pvnet/models/multimodal/validation/config_validation.py
```diff
@@ -0,0 +1,192 @@
...
+    validate_static_config,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def validate(
+    numpy_batch: NumpyBatch,
+    multimodal_config: dict,
+    input_data_config: dict,
```
AUdaltsova, line 26: This might be controversial, but I think these should be passed as config objects and dumped inside the function. This is less pretty, but I think would make the usage a lot smoother both by hand and if we want to integrate it as an automated check, and especially when used by hand (which I think is the current primary assumed use?) because the two require different tools to make them dicts and preparing everything just seems bulky to me. I think, ideally, it should be super seamless, maybe even the config paths should be unpacked automatically from the main config, which would actually ensure you're validating the same thing your model will end up using.

tests/test_valid_config_dict.yaml
felix-e-h-p: Ah purely for the tests

AUdaltsova: I can't find the usage of them in this PR, sorry, I might be looking it up weirdly - would you mind pointing me please? Though to be completely honest I think adding new parameters in the config just for testing is a bit dicey, it might result in tests not replicating actual behaviour they are supposed to be testing for... but maybe it's dealt with somehow here, I'm struggling to find the relevant bits, sorry!

pvnet/models/multimodal/validation/config_validation.py
felix-e-h-p: Definitely not controversial - agree completely here

pvnet/models/multimodal/validation/config_batch_validation.py
```diff
@@ -78,74 +78,88 @@ def get_modality_interval(
...
-                and secondary_modality_key is not None
-                and secondary_modality_key in input_data_config):
-            logger.debug(f"Using '{secondary_modality_key}' interval as fallback for sun.")
-            lookup_key = secondary_modality_key
-            error_context = f"fallback modality '{lookup_key}' for sun"
-            modality_config_dict = input_data_config.get(lookup_key)
+        if "nwp" not in input_data_config or not isinstance(input_data_config["nwp"], dict):
+                raise KeyError(
+                    "NWP section ('nwp') not found or is not a dictionary "
+                    "in input_data_config."
```
AUdaltsova, line 94: I'm continuing to split hairs but I think this is clear:
```suggestion
                    "in data configuration."
```
because input_data_config is an internal variable here and the actual config can be called something completely different

pvnet/models/multimodal/validation/config_batch_validation.py
```diff
@@ -51,22 +51,22 @@ def get_modality_interval(
...
+        primary_modality_key: The main key for the modality in `input_data_config`.
+            If this is "nwp", then `secondary_modality_key` is treated as the
+            specific NWP source name within the "nwp" section.
+        secondary_modality_key: An optional secondary key.
+            - If `primary_modality_key` is "nwp", this is REQUIRED and specifies
+              the NWP source (e.g., "ecmwf").
+            - If `primary_modality_key` is "sun", this can specify a fallback
+              modality (e.g., "gsp") for interval lookup if the "sun" config
+              is not found directly.
+            - Otherwise, typically None for non-NWP, non-sun direct lookups.
```
AUdaltsova, line 69: This is great, thank you!!

pvnet/models/multimodal/validation/config_batch_validation.py
```diff
@@ -78,74 +78,88 @@ def get_modality_interval(
...
+        error_context_desc = f"NWP source '{key_for_config}'"
+
+    modality_config_dict = config_section_to_search.get(key_for_config)
+
+    if modality_config_dict is None:
+        if primary_modality_key == "sun" and secondary_modality_key is not None:
+            logger.debug(
+                f"Config for '{primary_modality_key}' not found directly, "
+                f"attempting fallback to '{secondary_modality_key}' "
+                f"from top-level input_data_config."
```
AUdaltsova, line 108: Maybe I'm missing something here, I'm not sure what this is communicating?

pvnet/models/multimodal/validation/config_validation.py
```diff
@@ -88,105 +101,161 @@ def validate(
...
+                (k_sun for k_sun in possible_fallback_keys if k_sun in input_data_dict and
+                    isinstance(input_data_dict.get(k_sun), dict)),
+                None
+            )
+            if fallback_key_sun is not None:
+                try:
+                    sun_time_res_mins = get_modality_interval(
+                        input_data_dict, "sun", fallback_key_sun
+                    )
+                    logger.debug(
```
AUdaltsova, line 232: I feel like if there is already fallback modality handling within get_modality_interval we can just put sun + backup key through it right away and let it either handle it and raise appropriate warnings or fail. Is there a reason for these try/excepts I'm not seeing?

AUdaltsova: Also, strictly speaking, I think if sun_posiotion is not explicitly in data config it will not be in the batches either? At least for dsampler-created batches, because I think currently we expect sun position to be added on creation and only if sun pos config is present. We used to have this situation in datapipes where sun position would be calculated by them after batch creation, so you wouldn't have this data in the batch initially and it would be added later on, but we don't do that anymore.

Is there some mid-way models we are trying to accommodate for here? If not then I would say we should just fail hard if there is include_sun in config and no sun data/there is sun data but no data config, because that just means it's not the batch's config.

pvnet/models/multimodal/validation/config_batch_validation.py
```diff
@@ -0,0 +1,293 @@
...
+        final_expected_shape = expected_shape_with_batch + (1,)
+
+    for i in range(1, actual_ndim):
+        if data.shape[i] != final_expected_shape[i]:
+            raise ValueError(
+                f"'{data_key}' shape error at dimension {i} (0-indexed, after batch). "
+                f"Expected size {final_expected_shape[i]}, Got {data.shape[i]}. "
+                f"Full Expected Shape: {final_expected_shape}, Full Actual Shape: {data.shape}. "
+                f"(Time resolution for context: {time_resolution_minutes} mins)."
+            )
```
AUdaltsova, line 218: I still think this should have more information on the actual mismatch. Basically we want to transition from the current matmul errors we get and then have to hunt around for which dimension that could've come from to a very clear validation error that basically says: "Dimension mismatch! Expected width x from encoder configuration, got width y from batch size" or something similar. Given that we have to define the expected shape dim by dim before running this function anyway, we can e g supply a list of keys of what each dimension is so it's easier to put into the error message, and get something like "Dimension mismatch for {width} in {satellite data}: expected {x} from model config, found {y} in batch"

pvnet/models/multimodal/validation/config_validation.py
```diff
@@ -0,0 +1,261 @@
...
+    input_data_dict: dict = OmegaConf.to_container(
+        input_data_config, resolve=True, throw_on_missing=True
+    )
+
+    if not isinstance(expected_batch_size, int) or expected_batch_size <= 0:
+        raise ValueError(
+            f"expected_batch_size must be a positive integer, got {expected_batch_size}"
+        )
+    try:
+        validate_static_config(mm_config_dict)
```
AUdaltsova, line 55: I still think this should be renamed from static to model so it's clear what it's checking, but maybe others disagree?

AUdaltsova:
This is great work, thanks so much @felix-e-h-p for all this! Definitely getting there now

By the way, I think we should run model config validation on training init, and run batch validation for the 8 batches we run sanity checks on. Maybe that's a separate PR though

tests/test_valid_config_dict.yaml
AUdaltsova: Hi, sorry I don't think the changes for this went through, are they supposed to be here?

pvnet/models/multimodal/validation/config_batch_validation.py
```diff
@@ -78,99 +78,121 @@ def get_modality_interval(
...
             of the array, including the batch dimension as the first element.
         data_key: The key or name identifying the data array, used for context
             in error messages.
-        interval: The time interval associated with the data's time dimension,
+        time_resolution_minutes: The time resolution (in minutes) of the data,
             used for context in error messages.
         allow_ndim_plus_one: If True, allows the array's number of dimensions
-            to be one greater than defined by `expected_shape_with_batch`
-            (typically for an added channel dimension of size 1). Defaults to False.
+            # ...
```
AUdaltsova, line 168:
```suggestion
```

pvnet/models/multimodal/validation/config_batch_validation.py
```diff
@@ -78,99 +78,121 @@ def get_modality_interval(
...
         expected_shape_with_batch: A tuple representing the expected shape
             of the array, including the batch dimension as the first element.
         data_key: The key or name identifying the data array, used for context
             in error messages.
-        interval: The time interval associated with the data's time dimension,
+        time_resolution_minutes: The time resolution (in minutes) of the data,
             used for context in error messages.
         allow_ndim_plus_one: If True, allows the array's number of dimensions
-            to be one greater than defined by `expected_shape_with_batch`
-            (typically for an added channel dimension of size 1). Defaults to False.
```
AUdaltsova: I think this was supposed to stay? Am I reading this wrong?

pvnet/models/multimodal/validation/config_batch_validation.py
```diff
@@ -179,26 +201,44 @@ def validate_array_shape(
...
 
     actual_ndim = data.ndim
-
     if actual_ndim not in allowed_ndims:
         allowed_ndims_str = " or ".join(map(str, sorted(list(allowed_ndims))))
         raise ValueError(
-            f"'{data_key}' dimension error. Expected {allowed_ndims_str} dims, Got {actual_ndim}"
+            f"'{data_key}' dimension count error. "
+            f"Expected {allowed_ndims_str} dims, Got {actual_ndim}. "
+            f"(Time resolution for context: {time_resolution_minutes} mins)."
```
AUdaltsova, line 209:
```suggestion
```
Same here, time resolution doesn't affect ndims

pvnet/models/multimodal/validation/config_batch_validation.py
```diff
@@ -78,99 +78,121 @@ def get_modality_interval(
...
     if not isinstance(data, np.ndarray):
-        raise TypeError(f"'{data_key}' data must be NumPy array, found {type(data).__name__}")
+        raise TypeError(f"'{data_key}' data must be NumPy array, found {type(data).__name__}.")
     if data.ndim == 0:
         raise ValueError(f"'{data_key}' array is scalar.")
 
     actual_batch_size = data.shape[0]
     if actual_batch_size <= 0:
-        raise ValueError(f"'{data_key}' batch dimension has size <= 0: {actual_batch_size}")
+        raise ValueError(f"'{data_key}' batch dimension has size <= 0: {actual_batch_size}.")
```
AUdaltsova, line 185: Maybe I'm wrong, but I still think you can't really get a negative value out of ndarray shape?

pvnet/models/multimodal/validation/batch_content_validator.py
```diff
@@ -182,7 +181,7 @@ def validate_array_shape(
 
     actual_batch_size = data.shape[0]
     if actual_batch_size <= 0:
-        raise ValueError(f"'{data_key}' batch dimension has size <= 0: {actual_batch_size}.")
+        raise ValueError(f"'{data_key}' batch dimension has size = 0: {actual_batch_size}.")
```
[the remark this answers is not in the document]
felix-e-h-p: Just pretty much to validate that the batch dimension, i.e. first dimension of the array, is positive. Early / semantically focused error if the batch itself is potentially empty or invalid - as a prerequisite to further targeted checks really, such as whereby checking batch size matches exact positive number etc.

PR #369, opened 2025-04-29
Fix imports in save_samples.py

AUdaltsova:
# Pull Request

## Description

We've moved samples inside torch_datasets in  mob clean, so the import in save_samples.py needs to be updated

## How Has This Been Tested?

Creating samples with the updated script

- [x] Yes

### DIFF
--- a/scripts/save_samples.py
+++ b/scripts/save_samples.py
@@ -43,9 +43,9 @@ import warnings
 
 import dask
 import hydra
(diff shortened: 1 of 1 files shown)


Sukh-P:
Thanks for fixing this! I am happy for this to merged even though the test is failing as that seems like an xarray issue we need to address, are you okay to open a bug issue for that failing test?

PR #370, opened 2025-04-30
Update example config with correct DataModule path

AUdaltsova:
# Pull Request

## Description

- DataModule(s) are now imported directly from pvnet/data/__init__.py and target pvnet.data.datamodule.DataModule is no longer valid
- Added sun position config to the example configuration

## How Has This Been Tested?

Ran run.py with updated premade_batches.yaml as datamodule

- [X] Yes

### DIFF
--- a/configs.example/datamodule/configuration/example_configuration.yaml
+++ b/configs.example/datamodule/configuration/example_configuration.yaml
@@ -281,3 +281,8 @@ input_data:
       WV_073:
         mean: 0.62479186
         std: 0.12924142
+
+  solar_position:
+    interval_start_minutes: -60
+    interval_end_minutes: 480
+    time_resolution_minutes: 5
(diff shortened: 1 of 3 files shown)


Sukh-P (approved):

configs.example/datamodule/configuration/example_configuration.yaml
```diff
@@ -281,3 +281,8 @@ input_data:
       WV_073:
         mean: 0.62479186
         std: 0.12924142
+
+  solar_position:
+    interval_start_minutes: -60
+    interval_end_minutes: 480
+    time_resolution_minutes: 5
```
Sukh-P, line 288: Just a suggestion it might be slightly clearer as an example if the time resolution was 15 or 30 as that matches either the site or gsp example resolution, not a big deal though

Sukh-P:
Some comment as here https://github.com/openclimatefix/PVNet/pull/369#issuecomment-2842482758 so feel free to merge

configs.example/datamodule/configuration/example_configuration.yaml
AUdaltsova, line 288:
```suggestion
    time_resolution_minutes: 30
```

PR #371, opened 2025-04-30
Remove timezone for synthetic data in conftest.py

AUdaltsova:
# Pull Request

## Description

New `xarray` release doesn't allow passing datetimes with timezone to `.to_netcdf`. We don't use timezone typically so removing this from the synthetic data in conftests.


Sukh-P (approved):
Nice work!

PR #375, opened 2025-05-09
Explicitly supply mapping between the location ID and the embedding index

dfulu:
# Pull Request

## Description

Explicitly supply mapping between the location ID and the embedding index

Previously we had it hard coded so that the GSP ID or site ID used as the embedding index. This PR changes the behaviour so that a mapping between the location (GSP or site) ID and the embedding index can be used. 

This supports the move to train PVNet with the new GSP definitions (openclimatefix/uk-pvnet-app#289) and moves us closer to being able to use strings as the location ID rather than integers.

## Checklist:

### DIFF
--- a/configs.example/model/multimodal.yaml
+++ b/configs.example/model/multimodal.yaml
@@ -68,6 +68,13 @@ embedding_dim: 16
 include_sun: True
 include_gsp_yield_history: False
 
+# The mapping between the location IDs and their embedding indices
+label_mapping:
+  1: 1
+  5: 2
+  110: 3
+# ...
+
 #--------------------------------------------
 # Times
 #--------------------------------------------
--- a/configs.example/model/multimodal.yaml
+++ b/configs.example/model/multimodal.yaml
@@ -69,7 +69,7 @@ include_sun: True
 include_gsp_yield_history: False
 
 # The mapping between the location IDs and their embedding indices
-label_mapping:
+location_id_mapping:
   1: 1
   5: 2
   110: 3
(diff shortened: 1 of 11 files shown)


pvnet/models/multimodal/multimodal.py
```diff
@@ -289,6 +301,13 @@ def forward(self, x):
         if self.adapt_batches:
             x = self._adapt_batch(x)
 
+        if self.use_id_embedding:
+            id = torch.tensor(
```
peterdudfield, line 305: could you an a comment with an example, like `x['gsp_id] = 1` with location_id_mapping  = {1:0}, would give [0], 
or something like that, just so its slightly easier to read and know what soging on

configs.example/model/multimodal.yaml
```diff
@@ -68,6 +68,13 @@ embedding_dim: 16
 include_sun: True
 include_gsp_yield_history: False
 
+# The mapping between the location IDs and their embedding indices
```
peterdudfield, line 71: nice its stored in here! i like this

peterdudfield (approved):

pvnet/models/multimodal/multimodal.py
```diff
@@ -47,6 +47,7 @@ def __init__(
         include_gsp_yield_history: bool = True,
         include_sun: bool = True,
         include_time: bool = False,
+        location_id_mapping: Optional[dict[Any, int]] = None,
```
peterdudfield, line 50: should the default value be {i:i for i in range(1,318)}

dfulu: It could be, and I think that would maintain backwards compatibility with the models we've already trained. But I also don't think we should encourage not setting this parameter. But I'll put it in

dfulu: I've added it in, but I'm not sure this is the ideal solution.

With the new GSPs where the IDs go above 317 using this default will cause the model training to crash. And for the India models where we only have a couple of IDs this default wastes some parameters - though only a small percentage of the total number

PR #377, opened 2025-05-12
Strip back

dfulu:
# Pull Request

## Description

Remove lots of code that we no longer use

## Checklist:


Sukh-P (approved):
Looks great! Thanks so much for removing ~2000 LoC! 😅

Sukh-P:
I think as previously mentioned could possibly also strip out the [wind_multimodal.py](https://github.com/openclimatefix/PVNet/blob/main/configs.example/model/wind_multimodal.yaml) file, don't think it adds anything extra as an example

PR #306, opened 2025-01-22
Stop the _adapt_batch() from changing the batch in-place

adityasuthar:
# Pull Request

## Description

The _adapt_batch() method changes the batch in place. This is bad practice. We are updating the method so that it returns a new batch without changing the old one.

Fixes #210 

## How Has This Been Tested?

I tested by running the `python -m pytest tests` command. 
- [x] Yes

_If your changes affect data processing, have you plotted any changes? i.e. have you done a quick sanity check?_

- [ ] Yes

## Checklist:

- [x] My code follows [OCF's coding style guidelines](https://github.com/openclimatefix/.github/blob/main/coding_style.md)
- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked my code and corrected any misspellings

Issue #210, opened 2024-06-05
Stop the `_adapt_batch()` from changing the batch in-place


dfulu:
The `_adapt_batch()` method changes the batch in-place. This is bad practice and led to a hard to find [bug in pvnet_app](https://github.com/openclimatefix/pvnet_app/pull/83)

adityasuthar:
Hi @dfulu I have raised a pull request for this issue at #306

adityasuthar:
Hi @dfulu, just a gentle reminder that my PR for issue #306 is awaiting review.

Dakshbir:
@dfulu 
Can i work on this issue (if it hasn't been solved yet)?
I believe that i will be able to fix the issue

PR #379, opened 2025-05-13
Adapt batch fix

dfulu:
# Pull Request

## Description

There was an error in how we use the `_adapt_batch()`. Previously we used to edit the batches in-place but this was [flagged to change in this issue](https://github.com/openclimatefix/PVNet/issues/210) and fixed in [this PR](https://github.com/openclimatefix/PVNet/pull/306). 

When this the `_adapt_batch()` was updated this introduced an error where the batch was updated for the forward pass of the model, but not for calculating the MAE and for backprop. This meant that when training for an intraday model using batches which go out to day ahead - the model was actually being trained the predict the last 8 hours of the 36 hour window.

## Checklist:


peterdudfield (approved):
Aproved, could you either add a test for this? Or add a github issue to test this?

dfulu:
> Aproved, could you either add a test for this? Or add a github issue to test this?

There isn't really an obvious test that could have caught this. The mistake was in the slicing inside the training and validation steps. Those train/val steps were passing but just weren't correctly indexed internally. 

We'd need quite a big refactor to unit test this bit, or a really convoluted test to make sure the MAE values returned by the train/val steps were as expected. So I don't think this is worth testing

PR #382, opened 2025-05-16
save model configs to yaml format

dfulu:
# Pull Request

## Description

Currently we are saving our model configs to json format (although we put them into a file called `.yaml`). This has worked fine until now. However, the new `location_id_mapping` introduced in #375 poses a problems for the json format. When saving to json it changes the integer keys of the `location_id_mapping` to strings. This breaks the model config so that the model will run after loading. 

This PR changes the saving so that we use the `yaml` library to save the model config


peterdudfield (approved):
Looks good (if test pass)

PR #402, opened 2025-06-09
Push - reduced main function for config only

felix-e-h-p:
Configuration only validation - relating to PR https://github.com/openclimatefix/PVNet/pull/366


tests/conftest.py
```diff
@@ -89,11 +99,79 @@ def sat_data():
...
+    config_dict = configuration_object.model_dump(mode='python')
+    input_data_dict = config_dict.get("input_data")
+
+    if input_data_dict is None:
+         raise KeyError(f"Key 'input_data' not found in configuration loaded from {config_path}")
+
+    return input_data_dict
+
+
+def convert_pytorch_dict_to_numpy(pytorch_dict: dict[str, object]) -> dict[str, object]:
```
peterdudfield, line 145: Why do you need this in this PR?

tests/conftest.py
```diff
@@ -89,11 +99,79 @@ def sat_data():
...
+            f"Test configuration file not found: {config_path}. "
+            f"Ensure '{config_filename}' exists in the '{conftest_dir}' directory."
+        )
+
+    configuration_object: Configuration = load_yaml_configuration(config_path)
+    config_dict = configuration_object.model_dump(mode='python')
+    input_data_dict = config_dict.get("input_data")
+
+    if input_data_dict is None:
+         raise KeyError(f"Key 'input_data' not found in configuration loaded from {config_path}")
```
peterdudfield, line 140: this is a ficture, so I dont think you need to check if, just make sure it is

tests/conftest.py
```diff
@@ -89,11 +99,79 @@ def sat_data():
...
+    return cfg_dict
+
+
+@pytest.fixture
+def valid_input_data_config() -> Dict[str, Any]:
+    config_filename = "test_input_data_config.yaml"
+    conftest_dir = os.path.dirname(os.path.abspath(__file__))
+    config_path = os.path.join(conftest_dir, config_filename)
+
+    if not os.path.exists(config_path):
```
peterdudfield, line 129: remove, just make sure fixture works

tests/conftest.py
```diff
@@ -89,11 +99,79 @@ def sat_data():
...
+            numpy_dict[key] = value.detach().cpu().numpy()
+        elif isinstance(value, dict):
+            numpy_dict[key] = convert_pytorch_dict_to_numpy(value)
+        else:
+            numpy_dict[key] = value
+    return numpy_dict
+
+
+@pytest.fixture()
+def sample_numpy_batch() -> NumpyBatch:
```
peterdudfield, line 158: do you need this for this PR?

pvnet/models/multimodal/validation/config_structure_validator.py
```diff
@@ -0,0 +1,501 @@
...
+            f"found {type(value).__name__}."
+        )
+        # Non critical / critical type mismatches
+        if warn_on_type_mismatch:
+            logger.warning(message)
+        else:
+            raise TypeError(message)
+
+
+def _check_dict_section(
```
peterdudfield, line 63: do you need this, or could you just do it on type above?

pvnet/models/multimodal/validation/config_structure_validator.py
```diff
@@ -0,0 +1,501 @@
...
+
+
+def _check_dict_values_are_int(
+    data: dict[str, Any],
+    dict_name: str,
+) -> None:
+    """Check if all values in a dictionary are integers (logs warning on mismatch)."""
+    expected_type = int
+    for source, value in data.items():
+        if not isinstance(value, expected_type):
```
peterdudfield, line 317: reduce number of lines with 
```
if not isinstance(value, int):
```

pvnet/models/multimodal/validation/config_structure_validator.py
```diff
@@ -0,0 +1,501 @@
...
+              if "_target_" not in section_content:
+                  raise KeyError(
+                      f"Config section '{section_name}' is missing required "
+                      f"sub-key: '_target_'"
+                  )
+
+    return section_content
+
+
+def _check_time_parameter(
```
peterdudfield, line 117: what does this do mroe than _check_key

pvnet/models/multimodal/validation/config_structure_validator.py
```diff
@@ -0,0 +1,395 @@
...
+    Args:
+        cfg: The main configuration dictionary.
+        section_key: Key of the encoder section (e.g., 'sat_encoder', 'nwp_encoders_dict').
+        context: Context for error messages.
+        source_key: Sub-key if section_key points to a dict of encoders (e.g., NWP source).
+
+    Raises:
+        KeyError, TypeError, ValueError: If parameters are missing, wrong type, or not positive.
+    """
+    convnet_params = [
```
peterdudfield, line 225: im not sure this is right, becasue differnt encodes have different parameters

pvnet/models/multimodal/validation/config_structure_validator.py
felix-e-h-p: Suppose it can be integrated into the main function in the specific file, but not too sure, what do you reckon?

_check_dict_section helper pretty much performs a check for required _target_ sub-key and returns the dict section - safeguard really to ensure that any section meant to be an object has the relevant target key needed

pvnet/models/multimodal/validation/config_structure_validator.py
```diff
@@ -0,0 +1,297 @@
...
+
+    return encoder_config
+
+
+def validate_static_config(cfg: dict[str, Any]) -> None:
+    """Perform static validation of the multimodal configuration dictionary.
+
+    Checks presence, types, and basic constraints of core config parameters
+    and sections before checking batch data.
+
```
peterdudfield, line 211: Could you add to the description of what this function does for example
```
1. check quantiles
2. check keys
3. check Satellite
4. check NWP 
```

pvnet/models/multimodal/validation/config_structure_validator.py
```diff
@@ -0,0 +1,321 @@
...
+    # Check time param values are integers
+    _check_key(
+        cfg, "nwp_history_minutes", required=True, expected_type=dict, context=context
+    )
+    _check_key(
+        cfg, "nwp_forecast_minutes", required=True, expected_type=dict, context=context
+    )
+    _check_key(
+        cfg, "nwp_interval_minutes", required=True, expected_type=dict, context=context
+    )
```
AUdaltsova, line 68: This is technically optional by design for backwards compatibility - it defaults to 60 min resolution. I think we should encourage conscious use of this but it shouldn't fail hard if interval minutes are not there. Especially it should specifically allow for not all providers to be present, because I think it's ok to only specify this for non-hourly data - but maybe we want to be strict here, not sure.

pvnet/models/multimodal/validation/config_structure_validator.py
```diff
@@ -0,0 +1,321 @@
...
+    Args:
+        cfg: The main configuration dictionary.
+        nwp_section: The validated 'nwp_encoders_dict' section content.
+
+    Raises:
+        KeyError: If required NWP time/interval parameter dictionaries are missing.
+        TypeError: If NWP time/interval parameters are not dictionaries.
+        ValueError: If keys in time/interval dicts do not match NWP sources.
+    """
+    nwp_sources: list[str] = list(nwp_section.keys())
```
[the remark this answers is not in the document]
felix-e-h-p: Yeah precedes encoder checks - not sure with this one, I personally prefer it in the sense of focusing on consistency between sources and time params initially, and to avoid any potential coupling. What do you reckon?

pvnet/models/multimodal/validation/config_structure_validator.py
```diff
@@ -0,0 +1,321 @@
...
+                raise TypeError(
+                    f"Config for NWP source '{source}' in 'nwp_encoders_dict' "
+                    f"must be a dictionary, found "
+                    f"{type(source_specific_encoder_config).__name__}."
+                )
+
+            _check_key(source_specific_encoder_config, "_target_", required=True,
+                       context=f"Config for NWP source '{source}'")
+
+            if "required_parameters" in source_specific_encoder_config:
```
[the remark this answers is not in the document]
felix-e-h-p: NWP encoder for eg: to define its own specific set of params that must be present, be integers, and have positive values - likewise with sat

pvnet/models/multimodal/validation/config_structure_validator.py
```diff
@@ -0,0 +1,321 @@
...
+        _check_key(sat_section, "_target_", required=True, context="Section 'sat_encoder'")
+        _check_key(
+            cfg,
+            "sat_history_minutes",
+            required=True,
+            expected_type=int,
+            context="Config with 'sat_encoder'",
+        )
+        if "required_parameters" in sat_section:
+            for param in sat_section["required_parameters"]:
```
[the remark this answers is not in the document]
AUdaltsova: I feel like I'm reading this wrong. As far as I understand, this will look for a parameter `required_parameters` in sat encoder from the model config. Do we assume that from now on we will be adding a parameter like this to encoders? Unless we already do and there's an update I've missed?

Issue #374, opened 2025-05-08
Add sample creation data config


Sukh-P:
Currently we don't save the data config information related to how many samples and train/validation splits anywhere that is easily accessible, so we want to change this and save this information either on W&B or HuggingFace, W&B would be the preferred option for now but would want to make sure that run link is also available on HuggingFace.

Sukh-P:
Need to check if experiment_config in W&B always saves correctly found an instance where it didn't seem to

PR #406, opened 2025-07-02
Save datamodule config

Sukh-P:
# Pull Request

## Description

Changes to also save datamodule config when training from presaved samples, this will allow full reproducibility/visibility of config used in training. 

This is done in two steps:
1) The changes in `training.py` means that the datamodule config is saved in the callback folder
2) The rest of the changes mean that when the callback folder is pushed to HF the datamodule file is also uploaded 

Tested this by running the save samples script and the checkpoint to huggingface script. 

Fixes https://github.com/openclimatefix/PVNet/issues/374

### DIFF
--- a/pvnet/load_model.py
+++ b/pvnet/load_model.py
@@ -2,11 +2,12 @@
 
 import glob
 import os
+from typing import Any
 
 import hydra
 import torch
-from typing import Any
 from pyaml_env import parse_config
+
 from pvnet.models.ensemble import Ensemble
 from pvnet.models.multimodal.unimodal_teacher import Model as UMTModel
 


pvnet/load_model.py
```diff
@@ -61,11 +69,13 @@ def get_model_from_checkpoints(
...
         model_config = model_configs[0]
         model = models[0]
-        data_config = data_configs[0]
 
-    return model, model_config, data_config
+    data_config = data_configs[0]
+    datamodule_config = datamodule_configs[0]
+
+
+    return model, model_config, data_config, datamodule_config
```
peterdudfield, line 81: Would you be able to update the function docs, and say what the return object. It defiantly takes me a while to work out what all the different configs are right now.

pvnet/load_model.py
```diff
@@ -52,6 +53,13 @@ def get_model_from_checkpoints(
...
 
+        # check for datamodule config
+        datamodule_config = f"{path}/datamodule_config.yaml"
+        if os.path.isfile(datamodule_config):
+            datamodule_configs.append(datamodule_config)
+        else:
+            datamodule_configs.append(None)
+
         model_configs.append(model_config)
         models.append(model)
```
AUdaltsova, line 64: This is a nitpick and not strictly speaking part of this PR but it feels strange that all model loading and manipulation happens first (LN27-47) but the model and its config are only added after all data configs have been found and added (LN63-64). I think it would be easier to read if everything model-related was consecutive?

pvnet/load_model.py
Sukh-P: Yep could definitely be a bit clearer, added those now

pvnet/load_model.py
Sukh-P: Yeah good point, I have moved where these are added now so it's less disjointed

AUdaltsova (approved):
LGTM! Thanks for doing this

PR #414, opened 2025-07-14
Add epoch-consistent shuffling to validation streamed sample dataloader

dfulu:
# Pull Request

## Description

If we are streaming samples during training (which we hope to do using xarray-tensorstore openclimatefix/ocf-data-sampler#254) then our current implementation means that the validation samples are not shuffled. 

We often only use a subset of the validation samples since we set the `limit_val_batches` parameter in the lightning `Trainer` object. Since we aren't iterating through the entire validation set, this means we only from the start of the validation period rather than sampled throughout the period.

This PR adds a shuffle to the validation dataset, so that the set of validation samples used during training is sampled from the whole time period. The shuffle is done such that the exact same random selection of validation samples will be used on each epoch.

## Checklist:


Sukh-P (approved):
LGTM, thanks for adding this!

Issue #225, opened 2024-06-18
Remove the optimiser before uploading the model to huggingface


dfulu:
Due to momentum, the optimiser can take up a lot disk space compared to the model itself. We currently upload the full model, including the optimiser to huggingface and download that to the app. We should remove the optimiser before uploading to save on bandwidth and memory

PR #420, opened 2025-07-22
Separate the pytorch model from the lightning training module

dfulu:
# Pull Request

## Description

This PR introduces a split between the lightning training module and the pytorch model itself.

This allows us to upload only the model to huggingface rather than also uploading the optimisers and other paraphernalia connected to training the model. This solves #225 

I think this also introduces a cleaner split in the code-base between the code needed to run inference with the model and the code used to train and log results from training the model.

This is going to break backwards compatibility slightly with our already trained models. However, with some work it should be possible to import them into this new version. See the migration notes below

**TODO**

I still need to re-add the mutli-modal teacher, if we want to keep it. This will wait for a later PR into dev

**Extras**
- I have also taken the liberty of stripping out a couple of unused bits from the training routine. `timestep_intervals_to_plot` and `forecast_minutes_ignore` are no longer included
- Added some type hinting and a couple of clean-ups

**Testing** 

- All pytests pass
- I've trained a model with this new version and ran the checkpoint to huggingface script

----

**Migration**
The only change which needs to be made to the old models as of this PR is to delete the "optimizer" entry in the model config.

I have included a script which can be used to migrate the old models as stored on huggingface to this version of the library.

----

## Checklist:

### DIFF
--- a/scripts/migrate_old_model.py
+++ b/scripts/migrate_old_model.py
@@ -19,7 +19,7 @@ revision = "6feaa986a6bed3cc6c7961c6bf9e92fb15acca6a"
 # The local directory which will be downloaded to
 local_dir = "/home/jamesfulton/tmp/model_migration"
 
-# Whether to upload the migrated model back to the huggingface
+# Whether to upload the migrated model back to the huggingface repo
 upload = True
 
 # ------------------------------------------


pvnet/load_model.py
```diff
@@ -56,23 +54,23 @@ def get_model_from_checkpoints(
...
-        models.append(model)
+        # Extract the model from the lightning module
+        models.append(lightning_module.model)
+        model_configs.append(model_config["model"])
 
         # Check for data config
         data_config = f"{path}/{DATA_CONFIG_NAME}"
 
+        # Store the data config used for the model
+        data_config = f"{path}/{DATA_CONFIG_NAME}"
```
Sukh-P, line 67: Think this is a duplicate line of code and I am guessing what was meant was to just change the comment

pvnet/load_model.py
```diff
@@ -90,6 +88,7 @@ def get_model_from_checkpoints(
         model_config = model_configs[0]
         model = models[0]
 
+    # Assume if using an ensemble that the members were trained on the same input data
```
Sukh-P, line 91: Useful comment

pvnet/load_model.py
dfulu: Oops yes! I think I messed up the merge

scripts/migrate_old_model.py
```diff
@@ -0,0 +1,93 @@
...
+# USER SETTINGS
+
+# The model commit you want to update
+repo_id = "openclimatefix/pvnet_uk_region"
+revision = "6feaa986a6bed3cc6c7961c6bf9e92fb15acca6a"
+
+# The local directory which will be downloaded to
+local_dir = "/home/jamesfulton/tmp/model_migration"
+
+# Whether to upload the migrated model back to the huggingface
```
Sukh-P, line 22:
```suggestion
# Whether to upload the migrated model back to huggingface
```
or this?

```suggestion
# Whether to upload the migrated model back to the huggingface repo
```

Sukh-P (approved):
Really great work on all the improvements, definitely much cleaner this way!

scripts/migrate_old_model.py
```diff
@@ -0,0 +1,93 @@
...
+
+from pvnet.models.base_model import BaseModel
+from pvnet.utils import MODEL_CARD_NAME, MODEL_CONFIG_NAME
+
+# ------------------------------------------
+# USER SETTINGS
+
+# The model commit you want to update
+repo_id = "openclimatefix/pvnet_uk_region"
+revision = "6feaa986a6bed3cc6c7961c6bf9e92fb15acca6a"
```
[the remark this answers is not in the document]
dfulu: Sure

PR #422, opened 2025-07-23
Move/remove model cards and other minor tidy ups

dfulu:
# Pull Request

## Description

This PR does some rearranging of modules. 

- The `pvnet.models.utils` module is renamed for clarity
- The `model_cards` folder is moved up a level

One change for the user to be aware of is that we will no-longer version control the model cards inside PVNet. Instead we provide a template and the user can fill this out themselves. I think this is better now that PVNet is being used for multiple projects and keeping a deck of model cards, and keeping them up to date with input data etc is rather cumbersome

### DIFF
--- a/pvnet/model_cards/empty_model_card_template.md
+++ b/pvnet/model_cards/empty_model_card_template.md
@@ -22,7 +22,6 @@ This model class uses satellite data, and numerical weather predictions to forec
 
 # Training Details
 
-<!-- The data section should link to a Data Card, perhaps with a short stub of information on what the training data is all about as well as documentation related to data pre-processing or additional filtering. -->
 ## Data
 
 <!-- eg.
--- a/scripts/checkpoint_to_huggingface.py
+++ b/scripts/checkpoint_to_huggingface.py
@@ -1,12 +1,21 @@
 """Command line tool to push locally save model checkpoints to huggingface
 
-use:
+To use this script, you will need to write a custom model card. You can copy and fill out
+`pvnet/model_cards/empty_model_card_template.md` to get you started.
+
+These model cards should not be added to and version controlled in the repo since they are specific
+to each user.
+
+Then run using:
+
+```
 python checkpoint_to_huggingface.py "path/to/model/checkpoints" \
     --huggingface-repo="openclimatefix/pvnet_uk_region" \
     --wandb-repo="openclimatefix/pvnet2.1" \
-    --card-template-path="pvnet/model_cards/pv_uk_regional_model_card_template.md" \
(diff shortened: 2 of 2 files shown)


scripts/checkpoint_to_huggingface.py
```diff
@@ -4,7 +4,7 @@
 python checkpoint_to_huggingface.py "path/to/model/checkpoints" \
     --huggingface-repo="openclimatefix/pvnet_uk_region" \
     --wandb-repo="openclimatefix/pvnet2.1" \
-    --card-template-path="pvnet/models/model_cards/pv_uk_regional_model_card_template.md" \
+    --card-template-path="pvnet/model_cards/pv_uk_regional_model_card_template.md" \
```
Sukh-P, line 7: Should we change the example .md file name here to the example one now just to avoid confusion?

Sukh-P:
@dfulu thanks for making these changes, I agree it's probably more robust this way and the old method would get out of date quite quickly, one thing that is missing from the PR for me though is some explicit mention that the new expectation is to fill out the model card template with the relevant details now before the HF upload (as you mentioned in the PR description), this could be a comment in the checkpoint to hf script, or we could add it to the README or some other place?

scripts/checkpoint_to_huggingface.py
dfulu: I don't want to encourage anyone to use that example file. How about we just put in a placeholder? Like `"pvnet/models/model_cards/my_custom_model_card.md"`

pvnet/model_cards/empty_model_card_template.md
```diff
@@ -0,0 +1,61 @@
...
+-->
+
+- **Developed by:** openclimatefix
+- **Model type:** Fusion model
+- **Language(s) (NLP):** en
+- **License:** mit
+
+# Training Details
+
+<!-- The data section should link to a Data Card, perhaps with a short stub of information on what the training data is all about as well as documentation related to data pre-processing or additional filtering. -->
```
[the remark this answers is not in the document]
dfulu: Yeh sure. I'll remove the mention of it

dfulu:
> @dfulu thanks for making these changes, I agree it's probably more robust this way and the old method would get out of date quite quickly, one thing that is missing from the PR for me though is some explicit mention that the new expectation is to fill out the model card template with the relevant details now before the HF upload (as you mentioned in the PR description), this could be a comment in the checkpoint to hf script, or we could add it to the README or some other place?

I like the idea of putting it in the docstring of the checkpoint script. If you're happy I'll put it there?

Sukh-P (approved):

PR #425, opened 2025-07-23
Clean  `pvnet/models/base_model.py`

dfulu:
# Pull Request

## Description

- We can turn on `weights_only` in the torch load which is better practice
- Refactor the functions to clean the config so we aren't repeatedly loading and saving the config to disk 
- Refactor the function to download from huggingface so that we can download multiple files at once and avoid code duplication. Also remove some unused parameters to the function
- Remove unused dependencies
- Update ruff settings
- Some other minor cleaning


Sukh-P:
This looks good! Just checking in terms of testing this have all the new pathways been ran to check they are all fine?

dfulu:
> This looks good! Just checking in terms of testing have all the new pathways been ran to check they are all fine?

Yes, I tested everything that this touched

PR #427, opened 2025-07-25
`training` submodule | model renaming | type hinting

dfulu:
# Pull Request

First of all, although this looks like a lot of additions I think a huge amount of those are down to type-hinting and needing to split parameters over lines in order to get the type hints in. I haven't actually added any new code in this PR.

**Summary of this PR**
- Renaming inside the models dir. `pvnet.models.multimodal.multimodal.Model` ->  `pvnet.models.late_fusion.late_fusion.LateFusionModel` [1]
- Introduce a `pvnet.training` submodule and move training functionality into it [2]
- Lots of type hinting everywhere
- Some other minor clean ups

On [1], the name change was motivated because all the model we use and will ever use are multimodal, so the old name is not very useful. Hopefully this refactor will allow us to trial more more architectures, and so I wanted to open up more name-space. I thought this was good to do now considering we will need to run migrations of the old models anyway. I also updated the migration script to take this renaming into account

On [2], I just think this is cleaner. It allowed me to take the plotting function out of `pvnet.utils`, and the `training.py` module, and the lightning module code and gather them all togther. I think have them all gathered together under one module is cleaner

**Testing**
- All the pytests pass
- I ran the sample saving script locally, trained on the samples, and ran the checkpoint to hf script on the model
- I trained on streamed samples, and ran the checkpoint to hf script on the model
- I ran the migration script
- I have not run the backtest scripts - I'm sure if they are currently in working order anyway

\* Thinking about the above, its doesn't speak well of our pytests that lots of scripts need to be run locally to assess whether the core functionality works

## Checklist:


pvnet/models/base_model.py
```diff
@@ -12,6 +12,7 @@
 import yaml
 from huggingface_hub import ModelCard, ModelCardData, snapshot_download
 from huggingface_hub.hf_api import HfApi
+from ocf_data_sampler.torch_datasets.sample.base import TensorBatch
```
Sukh-P, line 15: Is this path to the `TensorBatch` definition correct? Thought they lived here currently https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/numpy_sample/common_types.py

README.md
```diff
@@ -8,39 +8,34 @@
...
 
+Some more structured notes on experiments we have performed with PVNet are 
+[here](https://docs.google.com/document/d/1VumDwWd8YAfvXbOtJEv3ZJm_FHQDzrKXR0jU9vnvGQg).
 
 
 ## Setup / Installation
 
 ```bash
-git clone https://github.com/openclimatefix/PVNet.git
+git clone git@github.com:openclimatefix/PVNet.git
```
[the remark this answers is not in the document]
Sukh-P: All good, they can still do either and agree most people using it are developers/contributors so having the default recommendation to SSH is fine, thanks

pvnet/models/base_model.py
Sukh-P: I think this works because the TensorBatch object is imported to that file but I think it's best practise to import from the definition file, that being said I think that maybe where the definitions live could be moved and the folder structure there changed a bit 😅 

So not a blocking issue but we should probably change this at some point

Issue #281, opened 2024-11-29
Adapt batches silent padding


Sukh-P:
Currently we have the `adapt_batches` configuration setting which allows the [_adapt_batches](https://github.com/openclimatefix/PVNet/blob/main/pvnet/models/multimodal/multimodal_base.py#L11) function to modify the batch shape, I have recently through a configuration mistake noticed if you supply a sample that includes NWPs with width and height smaller than the output size that it will pad the NWP data in the samples with 0 values, as per torchvisions [center_crop documentation ](https://pytorch.org/vision/main/generated/torchvision.transforms.functional.center_crop.html). 

This is done silently currently, I think we should either explicitly throw an exception for this case of input size smaller than requested output size as I don't see it as a case we want to support/part of the original intention of this function or at least throw a warning so it's more obvious this is happening, keen to hear thoughts on this, thanks! @peterdudfield @dfulu @AUdaltsova

PR #435, opened 2025-08-01
Batch validation

felix-e-h-p:
Function entitled validate_batch_against_config added to pvnet/utils.py - to compare the shapes in a batch against the model config directly.

Triggers as a one time check on the initial training batch - implemented by on_train_batch_start in BaseModel class.

Should fundamentally now provide solid context regarding the exact data source and specific dimension that is incorrect.


tests/conftest.py
```diff
@@ -362,6 +358,7 @@ def raw_late_fusion_model_kwargs(model_minutes_kwargs) -> dict:
         sat_history_minutes=30,
         nwp_history_minutes={"ukv": 120, "ecmwf": 120},
         nwp_forecast_minutes={"ukv": 480, "ecmwf": 480},
+        nwp_interval_minutes={"ukv": 60, "ecmwf": 60},
```
peterdudfield, line 361: only change, the rest is ruff

pvnet/models/base_model.py
```diff
@@ -508,4 +511,10 @@ def _quantiles_to_prediction(self, y_quantiles: torch.Tensor) -> torch.Tensor:
         """
         # y_quantiles Shape: batch_size, seq_length, num_quantiles
         idx = self.output_quantiles.index(0.5)
-        return y_quantiles[..., idx]
\ No newline at end of file
+        return y_quantiles[..., idx]
+
+    def on_train_batch_start(self, batch, batch_idx):
```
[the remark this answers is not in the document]
felix-e-h-p: Felt perhaps better than situated in train.py - but yeah, not entirely sure

pvnet/models/base_model.py
```diff
@@ -508,4 +511,10 @@ def _quantiles_to_prediction(self, y_quantiles: torch.Tensor) -> torch.Tensor:
...
+        """Hook to run before training step for a single batch."""
+        if not self._has_validated_batch:
+            validate_batch_against_config(batch, self.hparams)
+            self._has_validated_batch = True
```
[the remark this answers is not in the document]
peterdudfield: do you mean to change it to model_config=self, or should it be model_config=self.hparams?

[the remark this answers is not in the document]
felix-e-h-p: Think better / more robust as =self as of the moment

pvnet/utils.py
```diff
@@ -84,3 +85,117 @@ def print_config(
...
+        sat_interval_minutes: The time resolution of the satellite data in minutes.
+        gsp_interval_minutes: The time resolution of the GSP data in minutes.
+
+    Raises:
+        ValueError: If a tensor shape mismatches the expected shape derived from the config.
+    """
+    logger.info("Performing batch shape validation against model config.")
+    dim_names = ["batch", "time", "channels", "height", "width"]
+
+    if "nwp" in batch and "nwp_encoders_dict" in model_config:
```
dfulu, line 153: This doesn't catch the case where there is an NWP encoder but no NWP data in the batch. That would slip through this check without raising anything. I think catching that would be a good addition. And similar for the other checks below

dfulu:
Currently this doesn't take into account our `_adapt_batches()` method and it will fail if the batches are too big, even though the `_adapt_batches()` method could make them fit the model. 

I think that either we should should discuss getting rid of the `_adapt_batches()` method (which might be a big discussion given that its really useful for presaved samples and its used in the uk-pvnet-app to run many models at once) or these checks should account for it.

This also relates quite closely to #281

tests/test_end2end.py
```diff
@@ -1,13 +1,20 @@
 import lightning
-from pvnet.training.lightning_module import PVNetLightningModule
+
 from pvnet.optimizers import EmbAdamWReduceLROnPlateau
+from pvnet.training.lightning_module import PVNetLightningModule
 
-def test_model_trainer_fit(late_fusion_model, uk_streamed_datamodule):
+
+def test_model_trainer_fit(late_fusion_model_and_config, uk_streamed_datamodule):
```
dfulu, line 7: Is there a reason not to just use the `late_fusion_model` and `late_fusion_model_kwargs` fixtures we already have instead of make a new `late_fusion_model_and_config` fixture and unpack it later?

pvnet/training/lightning_module.py
```diff
@@ -230,6 +230,13 @@ def on_validation_epoch_start(self):
     def validation_step(self, batch: TensorBatch, batch_idx: int) -> None:
         """Run validation step"""
 
+        # Batch validation check only during sanity check phase
+        if self.trainer.sanity_checking:
```
dfulu, line 234: This should go in `on_validation_epoch_start` because otherwise the code could error out in `on_validation_epoch_start()` before it reaches this point

PR #437, opened 2025-08-04
Add seed to val shuffle for reproducibility

dfulu:
# Pull Request

## Description

With the streamed sample datamodule, we previously did not control the seeding in it. When training two different models which had the same global seed and same data config we could end up with completely different val samples. This fixes the issue and will make it easier to compare between models.

Additionally remove unused print_config param in example config

## Checklist:

### DIFF
--- a/pvnet/data/base_datamodule.py
+++ b/pvnet/data/base_datamodule.py
@@ -102,6 +102,7 @@ class BaseStreamedDataModule(LightningDataModule):
         persistent_workers: bool = False,
         train_period: list[str | None] = [None, None],
         val_period: list[str | None] = [None, None],
+        seed: int | None = None,
 
     ):
         """Base Datamodule for streaming samples.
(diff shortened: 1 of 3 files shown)


pvnet/data/base_datamodule.py
```diff
@@ -142,11 +145,16 @@ def setup(self, stage: str | None = None):
...
         if stage == "fit":
+
             # Prepare the train dataset
             self.train_dataset = self._get_streamed_samples_dataset(*self.train_period)
 
-            # Prepare and pre-shuffle the val dataset
+            # Prepare and pre-shuffle the val dataset and set seed for reproducibility
             val_dataset = self._get_streamed_samples_dataset(*self.val_period)
+
+            if self.seed is not None:
```
peterdudfield, line 155: worthing adding a log.info("Setting seed to X")?

PR #444, opened 2025-08-19
Update site history

Sukh-P:
# Pull Request

## Description

These changes change the amount of historical site data is passed into PVNet, previously with `x["site"][:, : self.history_len + 1]` the t0 time would be included as part of the data sent to the model if for example the `history_minutes` was set to the same amount as the absolute value of the interval start if in the past, e.g.:

With the +1 in the indexing if `interval_start` was -60 and `interval_mintues` is 15 and `history_minutes` is 60 then it would take the first 5 timestamps (60/15 + 1) which would be relative to the t0 time [-60, ...., 0] where the last value is t0, we don't have t0 at inference time so it is better to not include this so the +1 is removed, this is more inline with how the GSP logic in the repo works as well for passing history

## Checklist:

### DIFF
--- a/pvnet/models/late_fusion/late_fusion.py
+++ b/pvnet/models/late_fusion/late_fusion.py
@@ -221,7 +221,7 @@ class LateFusionModel(BaseModel):
             assert pv_history_minutes is not None
 
             self.pv_encoder = pv_encoder(
-                sequence_length=pv_history_minutes // pv_interval_minutes + 1,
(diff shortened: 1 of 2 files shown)


dfulu (approved):

Sukh-P:
Not merging this in just yet to discuss the comments above in the team, also if merged in this should be flagged that this is not backwards compatible for already trained site models

PR #460, opened 2025-09-02
Add adapt batch check

Sukh-P:
# Pull Request

## Description

Change adds a check to see if `adapt_batches` has been set to `True` first before adapting batches in train/validation loop, prevents unnecessary operations being called when `adapt_batches` hasn't been set and stops some edge cases where forward pass data and backward pass data diverge (can happen with some experiment/hacking around setups). 

## Checklist:


dfulu (approved):
Looks good! I think we should really push towards getting rid of `_adapt_batch()` altogether, but thats for another issue

PR #461, opened 2025-09-08
Batch validation

felix-e-h-p:


pvnet/utils.py
```diff
@@ -79,10 +80,99 @@ def print_config(
...
+    batch: dict, 
+    model_config, 
+    sat_interval_minutes: int = 5, 
+    gsp_interval_minutes: int = 30, 
+    site_interval_minutes: int = 30
+) -> None:
+    """Validates tensor shapes in batch against model configuration."""
+    logger.info("Performing batch shape validation against model config.")
+    
+    if isinstance(model_config, (dict, DictConfig)):
```
dfulu, line 134: This should go. We should be more specific about what is passed into this function. We shouldn't be passing `dict` or `DictConfig`

pvnet/utils.py
```diff
@@ -1,11 +1,16 @@
...
 import logging
+from typing import TYPE_CHECKING
 
 import rich.syntax
 import rich.tree
+import torch
 from lightning.pytorch.utilities import rank_zero_only
 from omegaconf import DictConfig, OmegaConf
 
+if TYPE_CHECKING:
```
dfulu, line 11: This is cool. I didn't know you could do this

pvnet/utils.py
```diff
@@ -79,10 +84,95 @@ def print_config(
...
+                    actual_shape = tuple(nwp_data['nwp'].shape)
+                    raise ValueError(
+                        f"NWP.{source} shape mismatch: expected {expected}, got {actual_shape}"
+                    )
+
+    # Satellite validation
+    if hasattr(model_config, 'sat_encoder') and "sat" in batch:
+        enc = model_config.sat_encoder
+        expected_channels = enc.in_channels
+        if hasattr(model_config, "add_image_embedding_channel") and \
```
dfulu, line 155: Same as above. This `hasattr()` is always be true in our codebase

pvnet/utils.py
```diff
@@ -79,10 +84,95 @@ def print_config(
...
+                f"Expected: {expected_shape}, Actual: {actual_shape}."
+            )
+
+
+def validate_batch_against_config(
+    batch: dict, 
+    model_config: "BaseModel",
+    sat_interval_minutes: int = 5, 
+    gsp_interval_minutes: int = 30, 
+    site_interval_minutes: int = 30
```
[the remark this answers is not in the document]
dfulu: Wait, yes they are. My bad

pvnet/utils.py
```diff
@@ -79,10 +84,95 @@ def print_config(
...
+            actual_shape = tuple(batch['sat'].shape)
+            raise ValueError(f"Satellite shape mismatch: expected {expected}, got {actual_shape}")
+
+    # GSP/Site validation
+    target_configs = [("gsp", gsp_interval_minutes), ("site", site_interval_minutes)]
+    for key, interval in target_configs:
+        if (key in batch and hasattr(model_config, 'history_minutes') 
+            and hasattr(model_config, 'forecast_minutes')):
+            total_minutes = model_config.history_minutes + model_config.forecast_minutes
+            expected_len = total_minutes // interval + 1
```
dfulu, line 170: Could we not use [these model properties](https://github.com/openclimatefix/pvnet/blob/main/pvnet/models/base_model.py#L419-L420) and skip having to pass in `[gsp/site]_interval_minutes`?

pvnet/utils.py
```diff
@@ -79,10 +84,95 @@ def print_config(
...
+        expected = (batch["sat"].shape[0], enc.sequence_length, expected_channels, 
+                enc.image_size_pixels, enc.image_size_pixels)
+        if tuple(batch["sat"].shape) != expected:
+            actual_shape = tuple(batch['sat'].shape)
+            raise ValueError(f"Satellite shape mismatch: expected {expected}, got {actual_shape}")
+
+    # GSP/Site validation
+    target_configs = [("gsp", gsp_interval_minutes), ("site", site_interval_minutes)]
+    for key, interval in target_configs:
+        if (key in batch and hasattr(model_config, 'history_minutes')
```
dfulu, line 167: The `BaseModel` class always has the `history_minutes` property, so the `hasattr()` clause could go

pvnet/training/lightning_module.py
```diff
@@ -208,6 +207,17 @@ def on_validation_epoch_start(self):
 
             batch = collate_fn([val_dataset[i] for i in idxs])
             batch = self.transfer_batch_to_device(batch, self.device, dataloader_idx=0)
+
+            # if "satellite_actual" in batch:
```
dfulu, line 211: Can you get rid of these commented out lines

PR #471, opened 2025-09-18
Removed saved samples - updated README relatively also

felix-e-h-p:
- save_samples.py and save_concurrent_samples.py removed
- Related references removed
- ReadMe updated accordingly (not sure RE format)


dfulu (approved):
Looks good. Thanks!

PR #483, opened 2025-09-24
Remove presaved samples support

dfulu:
# Pull Request

## Description

This PR removes support for training from pre-saved samples

- The presaved samples Datasets and DataModules are removed along with their tests
- An if-clause for training on presaved samples is removed from `train.py`
- The remaining DataModules are renamed
- Combined and unnested the datamodule files since they are quite simple now

**Extra thoughts**
This leave the `test_datamodule.py` file very light. We just haven't been testing the streamed datamodules very much. We should probably add more tests, but I think that should be a different issue to this. This is about removing the presaved functionality


pvnet/datamodule.py
```diff
@@ -16,86 +17,7 @@ def collate_fn(samples: list[NumpySample]) -> TensorBatch:
...
-
-    def val_dataloader(self) -> DataLoader:
-        """Construct val dataloader"""
-        dataset = self._get_premade_samples_dataset("val")
-        return DataLoader(dataset, shuffle=False, **self._common_dataloader_kwargs)
-
-
-class BaseStreamedDataModule(LightningDataModule):
+class BaseDataModule(LightningDataModule):
     """Base Datamodule which streams samples using a sampler for ocf-data-sampler."""
```
Sukh-P, line 21: Small typo we can tidy I think 
```suggestion
    """Base Datamodule which streams samples using a sampler from ocf-data-sampler."""
```

Issue #432, opened 2025-07-30
Test core scripts


dfulu:
Currently our tests do not cover some of the core functionality of this library. We care if these scripts break but we don't currently test for it

We should test

- [ ] The backtest scripts
- [ ] The save checkpoint to huggingface script
   - Though the main method used in this script is tested already
- [x] Also the main `train()` function which is wrapped in the run.py script
   - Previously I had thought that the wandb logger might block testing the `train()` function, but I think it should work by setting the `offline=True` in the wandb logger

dfulu:
We decided that these scripts were not crucial and do not need to be tested

PR #491, opened 2025-10-10
Train related tests

felix-e-h-p:
Relating to https://github.com/openclimatefix/pvnet/issues/432

- Two tests for train() call in run.py
- Each builds minimal config using new fixtures (in file) with W&B in offline mode and checkpointing
- test_train_site trains with SitesDataModule (15-min)
- test_train_pv trains with UKRegionalDataModule (30-min)

Also updated relative to data sampler modifications

### DIFF
--- a/tests/training/test_train.py
+++ b/tests/training/test_train.py
@@ -9,7 +9,7 @@ from pvnet.training.train import train as pvnet_train
 
 
 @pytest.fixture()
-def wandb_offline(session_tmp_path) -> str:
+def wandb_save_dir(session_tmp_path) -> str:
     """Return W&B save dir under session temp path."""
     save_dir = str(session_tmp_path / "wandb")
     return save_dir
@@ -31,13 +31,13 @@ def trainer_cfg_cpu() -> dict:
 
 
 @pytest.fixture()
-def logger_cfg(wandb_offline: str) -> dict:
+def logger_cfg(wandb_save_dir: str) -> dict:
     """W&B logger config."""
     return {
         "wandb": {
             "_target_": "lightning.pytorch.loggers.wandb.WandbLogger",
             "project": "pvnet-tests",
-            "save_dir": wandb_offline,
+            "save_dir": wandb_save_dir,
             "offline": True,
             "name": "train-offline-integration",
             "log_model": False,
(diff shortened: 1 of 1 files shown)


tests/training/test_train.py
```diff
@@ -0,0 +1,162 @@
...
+from omegaconf import DictConfig
+
+from pvnet.training.train import train as pvnet_train
+
+
+@pytest.fixture()
+def wandb_offline_env(monkeypatch, session_tmp_path):
+    """Put W&B offline, quiet; force CPU."""
+    save_dir = str(session_tmp_path / "wandb")
+    monkeypatch.setenv("WANDB_SILENT", "true")
```
dfulu: We should avoid using environmental variables if we can. It adds complexity to the tests. Are these two variables needed for the test or do they clean up the test outputs or something?

dfulu: @felix-e-h-p thanks for the updates, I'm guessing the env vars weren't needed?

You might want to drop`monkeypatch` now that its no longer used in this fixture and maybe update the fixture name and doc to reflect what it does now

felix-e-h-p: Ahh yeah! Sorry - well spotted. Will push now, finalised you reckon?

tests/training/test_train.py
```diff
@@ -0,0 +1,161 @@
...
+def wandb_offline_env(monkeypatch, session_tmp_path):
+    """Put W&B offline, quiet; force CPU."""
+    save_dir = str(session_tmp_path / "wandb")
+    return save_dir
+
+
+@pytest.fixture()
+def trainer_cfg_cpu():
+    """Tiny CPU-only Trainer config."""
+    return dict(
```
dfulu: You've used the `{}` dictionary constructor in the other fixtures. It would be cleaner to standardise and use that here too

tests/training/test_train.py
```diff
@@ -0,0 +1,161 @@
...
+from pathlib import Path
+
+import pytest
+from omegaconf import DictConfig
+
+from pvnet.training.train import train as pvnet_train
+
+
+@pytest.fixture()
+def wandb_offline_env(monkeypatch, session_tmp_path):
```
dfulu: In the fixtures in the `tests/conftest.py` we include basic type hints of the return type. It would be good to add that to your fixtures here too

tests/training/test_train.py
```diff
@@ -0,0 +1,161 @@
...
+from pathlib import Path
+
+import pytest
+from omegaconf import DictConfig
+
+from pvnet.training.train import train as pvnet_train
+
+
+@pytest.fixture()
+def wandb_offline(session_tmp_path) -> str:
```
dfulu: This might be better to name this fixture something like `wandb_save_dir`?

PR #496, opened 2025-11-12
Generalised data sampler support

Sukh-P:
# Pull Request

## Description

 Changes for PVNet to support related changes in `ocf-data-sampler` to have one dataset/schema https://github.com/openclimatefix/ocf-data-sampler/pull/375

The changes here include:
- Updates to the target key and the input key name in the late_fusion.py file
    - Removal of the target key parameter in this file
- Changes and simplification of the datamodule
- Updating tests
- Update to the migration script for huggingface models
- Some linting changes after applying ruff

## How has this been tested?

Through unit tests and also running training with these changes and the related ocf-data-sampler changes, tracking results here https://wandb.ai/openclimatefix/test_new_data_sampler to check no large changes have occured to outputs

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ readme = {file="README.md", content-type="text/markdown"}
 requires-python = ">=3.11,<3.14"
 
 dependencies = [
-    "ocf-data-sampler@git+https://github.com/openclimatefix/ocf-data-sampler.git@dev-one-torch-dataset",
+    "ocf-data-sampler>=0.6.0",
     "numpy",
     "pandas",
     "matplotlib",


configs.example/model/late_fusion.yaml
```diff
@@ -68,8 +68,7 @@ model:
 
   embedding_dim: 16
   include_sun: True
-  include_gsp_yield_history: False
-  include_site_yield_history: False
+  include_generation_yield_history: False
```
dfulu: Since we're doing breaking changes here, I wonder if this might be a chance to reconsider this parameter name. I think `include_generation_history` might be better?

pvnet/models/base_model.py
```diff
@@ -399,22 +398,21 @@ def __init__(
...
             output_quantiles: A list of float (0.0, 1.0) quantiles to predict values for. If set to
                 None the output is a single value.
-            target_key: The key of the target variable in the batch
             interval_minutes: The interval in minutes between each timestep in the data
         """
         super().__init__()
 
-        self._target_key = target_key
+        # The key of the target variable in the batch
+        self._target_key = "generation"
```
dfulu: Should we get rid of this altogether and hard-code this where its needed? I'm actually not sure what the better practice would be in this instance. I can't think of a time where we'd want to change `self._target_key` to something else?

PR #501, opened 2026-01-06
Add a backtest script for models trained with ocf-data-sampler v1

AUdaltsova:
Previous backtest scripts are compatible with models trained with either GSP or site pipeline. The new script is based on the old GSP script and can be used to backtest models trained with ocf-data-sampler v1.

This script is still set-up in a way that expects a regional and a summation model. It might have to be modified further for site-specific models.

I have tested it with regional+summation models for Netherlands.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -150,3 +150,9 @@ dmypy.json
 
 # vim
 *swp
+
+# uv
+uv.lock
+
+# vscode
+.vscode/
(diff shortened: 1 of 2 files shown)


dfulu:
@AUdaltsova would you be happy for this to be closed now that there is a new backtest script?

AUdaltsova:
Yeah sure, sorry completely forgot this existed. Thanks!

PR #520, opened 2026-04-21
Make val and train periods more flexible

dfulu:
# Pull Request

## Description

Adds support for openclimatefix/ocf-data-sampler#412

This allows the training and validation datasets to be composed of different periods instead of our current restricted approach where the val period is always after the train period

## Checklist:

### DIFF
--- a/configs.example/datamodule/streamed_samples.yaml
+++ b/configs.example/datamodule/streamed_samples.yaml
@@ -8,12 +8,12 @@ prefetch_factor: 2
 persistent_workers: false
 batch_size: 8
 
-train_period:
-  - null
-  - "2022-05-07"
-val_period:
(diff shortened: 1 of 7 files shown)


Sukh-P (approved):
Looked good, thanks for adding this in!

One small thing but not blocking this change I think in the backtest scripts it would be good to update the type hints / put the date inputs into lists just so they're compatible with this change

dfulu:
> Looked good, thanks for adding this in!
> 
> One small thing but not blocking this change I think in the backtest scripts it would be good to update the type hints / put the date inputs into lists just so they're compatible with this change

I'm going to do this in a separate PR since I'm revamping the backtest script(s) anyway

PR #521, opened 2026-04-27
Add new backtests script

dfulu:
# Pull Request

## Description

- This PR adds a new backtest script (based on `backtest_uk_gsp.py`) which is compatible with the new combined data-sampler and replaces the two old scripts. 
- This PR makes the script more easy to configure through command line arguments and an input data config. 
- It also changes the saving method so that forecasts are appended to the output zarr as it progresses. 
- Also some refactoring w.r.t. `backtest_uk_gsp.py`
- I've removed the `backtest_sites.py script since this should replace the need for it and I believe that script is out of date now

This script now unfortunately uses the`StreamedDataset` class from `pvnet_summation`. This isn't ideal since the pvnet is strictly not supposed to import from `pvnet_summation` to avoid circular dependencies. Since this is a script rather than the core library I've decided to break that rule here. But It indicates that `pvnet_summation`'s `StreamedDataset` is perhaps doing too much and should be moved to `ocf-data-sampler` or perhaps this script should be moved to `pvnet_summation` or some bigger restructure into a `pvnet`+`pvnet_summation` monorepo would be appropriate once the future of `pvnet_summation` is decided. Either way I don't think this should be part of this PR - just a note

## How Has This Been Tested?

I've ran this script locally and checked the outputs manually. This was only for UK PV though I've tried to make it general enough that it should be able to run anything which uses data-sampler v1

## Checklist:


scripts/backtest.py
```diff
@@ -0,0 +1,477 @@
...
+
+    # Replace the generation data path
+    config["input_data"]["generation"]["zarr_path"] =  data_paths["generation"]
+
+    # Replace satellite data path if using it
+    if "satellite" in config["input_data"]:
+        if config["input_data"]["satellite"]["zarr_path"] != "":
+            config["input_data"]["satellite"]["zarr_path"] = data_paths["satellite"]
+
+    # NWP is nested so much be treated separately
```
AUdaltsova:
```suggestion
    # NWP is nested so must be treated separately
```

scripts/backtest.py
```diff
@@ -0,0 +1,477 @@
...
+        config: The data config
+        data_paths: The paths to the input data
+    """
+
+    # Replace the generation data path
+    config["input_data"]["generation"]["zarr_path"] =  data_paths["generation"]
+
+    # Replace satellite data path if using it
+    if "satellite" in config["input_data"]:
+        if config["input_data"]["satellite"]["zarr_path"] != "":
```
AUdaltsova: Do we have models with configs storing empty paths? Sorry if this is obvious, I've just not seen one

scripts/backtest.py
```diff
@@ -0,0 +1,477 @@
...
+    Args:
+        config: The data config
+    """
+    if "satellite" in config["input_data"]:
+
+        satellite_config = config["input_data"]["satellite"]
+
+        if satellite_config["zarr_path"] != "":
+            satellite_config["dropout_timedeltas_minutes"] = []
+            satellite_config["dropout_fraction"] = 0
```
AUdaltsova, line 116: Can we add a warning to this maybe? It's just something I ran into with NL, I didn't expect the script to do that and it took a bit for me to find what was going on

scripts/backtest.py
```diff
@@ -0,0 +1,477 @@
...
+# FUNCTIONS
+
+_model_mismatch_msg = (
+    "The PVNet version running in this app is {}/{}. The summation model running in this app was "
+    "trained on outputs from PVNet version {}/{}. Combining these models may lead to an error if "
+    "the shape of PVNet output doesn't match the expected shape of the summation model. Combining "
+    "may lead to unreliable results even if the shapes match."
+)
+
+def populate_config_with_data_data_filepaths(config: dict, data_paths: dict) -> dict:
```
AUdaltsova: Is there a reason for "data" being twice in there? I feel like I'm not reading it right maybe

scripts/backtest.py
```diff
@@ -0,0 +1,477 @@
...
+                "backtest_t0": t0,
+                "backtest_national_capacity": total_capacity,
+            }
+        )
+
+        return sample
+
+
+class Forecaster:
+    """Class for making and solar forecasts for all regions and optionally the national total"""
```
AUdaltsova:
```suggestion
    """Class for making solar forecasts for all regions and optionally the national total"""
```

scripts/backtest.py
```diff
@@ -0,0 +1,477 @@
...
+            da_abs_national = da_normed_national.clip(0, None) * national_capacity
+
+            if self.min_solar_elevation is not None:
+                # Apply sundown mask - All regions must be masked to mask national
+                da_abs_national = da_abs_national.where(
+                        ~da_sundown_mask.all(dim="location_id"), 
+                        other=0.0,
+                )
+
+            # Convert to Dataset and add attrs about the models used
```
AUdaltsova: I think this was supposed to be on L285?

AUdaltsova (commented):
Looks great, thanks for adding this! Left some nitpicks, looks good otherwise

scripts/backtest.py
dfulu: Yeh good point. I think this code block has been around for a long time. Let's kick it

scripts/backtest.py
dfulu: Sure

scripts/backtest.py
dfulu: What you've never heard of data-data-filepaths!?

dfulu:
> Looks great, thanks for adding this! Left some nitpicks, looks good otherwise

Thanks for catching all these @AUdaltsova! I've fixed those up and sneaked in a couple of extra clean ups, docs, and metadata pieces. But nothing major

scripts/backtest.py
```diff
@@ -280,16 +296,20 @@ def predict(self, sample: dict) -> xr.Dataset:
...
         da_abs.attrs.update(
             {
                 "pvnet_model_name": self.pvnet_model_name,
                 "pvnet_model_version": self.pvnet_model_version or "none",
                 "summation_model_name": self.summation_model_name or "none",
                 "summation_model_version": self.summation_model_version or "none",
+                "min_solar_elevation": (
+                    self.min_solar_elevation if self.min_solar_elevation is not None else "none",
+                ),
+                "date_of_backtest": pd.Timestamp.now().isoformat(),
```
AUdaltsova, line 312: Oh nice I like that!

scripts/backtest.py
```diff
@@ -405,72 +425,74 @@ def main(
     else:
         device = torch.device(device_name)
 
-    # Set up output dir for temporary files during backtest 
-    os.makedirs(output_dir, exist_ok=False)
+    if os.path.exists(output_zarr_path):
+        raise FileExistsError(
+            f"Output zarr path {output_zarr_path} already exists. Please choose a different  "
+            "output_dir or remove the existing zarr."
```
AUdaltsova: Is this still `output_dir`? or should it be `output_zarr_path` now?

dfulu: Oh yes good point! I missed that

AUdaltsova (approved):
Looks great, thanks so much for cleaning all this & putting it up!

PR #522, opened 2026-05-11
Use global step in all plots

### LINKED ISSUE #511: Having validation sample plots indexed by epochs
Just wanted to discuss whether it would be useful to try and get the slider/index for validation sample plots to which epoch it related to? (if this is possible) 

e.g. we currently have: 

Image

Where the slider index is based on how many times the wandb.log() has been called (I think this is default behaviour) but it's a bit fiddly then to align which samples have been created which which epoch version of the model

dfulu:
# Pull Request

## Description

This updates the forecast samples plots saved to wandb so the slider indexes on `global_step`.

Fixes #511 

## How Has This Been Tested?

I ran a training job - you can see the outputs here: https://wandb.ai/openclimatefix/wandb_plot_test

## Checklist:

### DIFF
--- a/pvnet/training/lightning_module.py
+++ b/pvnet/training/lightning_module.py
@@ -206,23 +206,25 @@ class PVNetLightningModule(pl.LightningModule):
             if self.trainer.sanity_checking and plot_num == 0:
                 validate_batch_against_config(batch=batch, model=self.model)
 
-            with torch.no_grad():
+            # Save example forecast plots via logger
+            if self.logger:
                 y_hat = self.model(batch)
 
-            fig = plot_sample_forecasts(
-                batch,
-                y_hat,
-                quantiles=self.model.output_quantiles,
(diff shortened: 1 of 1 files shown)


pvnet/training/lightning_module.py
```diff
@@ -206,23 +206,25 @@ def on_validation_epoch_start(self):
             if self.trainer.sanity_checking and plot_num == 0:
                 validate_batch_against_config(batch=batch, model=self.model)
 
-            with torch.no_grad():
```
dfulu, line 209: This lightning hook already sets the model to `no_grad()` so we don't need this

dfulu: Admittedly this was a little cleanup unrelated to the main changes

Sukh-P (approved):
Nice, will definitely make those graphs easier to compare

PR #525, opened 2026-05-18
Feature/crop history from sample plots

### LINKED ISSUE #523: Crop generation history and dropped values on wandb example plots
Currently, if a sample contains a lot of generation history and/or some dropped out values, the example plot will be really squashed to accommodate that and the prediction part of the plot can be quite hard to read. We should chop off both and just plot the prediction.

Sukh-P:
# Pull Request

## Description

Crops the sample forecast validation plots to just the forecasted period by removing the history. 

Note since dropped values (sometimes set to -1) are part of the history data these would be excluded too. 

## Testing
This has been tested by running an [example training run](https://wandb.ai/openclimatefix/gdm-gen/runs/oxiwnbhz) and looking at the sample plots 

Fixes #523

### DIFF
--- a/pvnet/training/plots.py
+++ b/pvnet/training/plots.py
@@ -32,7 +32,7 @@ def plot_sample_forecasts(
 
     y = batch[key_to_plot].cpu().numpy()
     y_hat = y_hat.cpu().numpy()
-    n_forecast = y_hat.shape[1]
+    forecast_length = y_hat.shape[1]
     ids = batch["location_id"].cpu().numpy().squeeze()
     times_utc = pd.to_datetime(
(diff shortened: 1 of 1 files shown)


Sukh-P:
@AUdaltsova I've taken the simple approach here and just cropped all of the history as this solves both of the initial issues you mentioned in #523 but I was just thinking would it still be useful to see the last few points of history and just crop the y axis so range so negative values below a certain amount aren't shown? 

Let me know what you think and whether a slightly more complex fix is worth it, thanks

AUdaltsova (approved):
LGTM, thanks! Like the variable `forecast_times` as well instead of all the `times_utc` indexing, makes it a lot clearer in the long run imo.

Issue #528, opened 2026-06-04
Change default drop_last=False to True


felix-e-h-p:
This will remove last incomplete batch and guarantees every batch has exactly specified batch_size.

This prevents BatchNorm from ever seeing batch size=1 and related errors when training.

https://github.com/openclimatefix/pvnet/blob/main/pvnet/datamodule.py#L69

dfulu:
Solved by #529

PR #529, opened 2026-06-05
Set drop_last to True in DataLoader

felix-e-h-p:
Refer to: https://github.com/openclimatefix/pvnet/issues/528

### DIFF
--- a/pvnet/datamodule.py
+++ b/pvnet/datamodule.py
@@ -66,7 +66,7 @@ class PVNetDataModule(LightningDataModule):
             num_workers=num_workers,
             collate_fn=collate_fn,
             pin_memory=pin_memory,
-            drop_last=True,
+            drop_last=False,
             timeout=0,
             worker_init_fn=None,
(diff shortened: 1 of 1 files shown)


pvnet/datamodule.py
```diff
@@ -121,8 +120,17 @@ def _get_dataset(self, time_periods: list[tuple[str | None, str | None]]) -> PVN
...
+            **self._common_dataloader_kwargs
+        )
 
     def val_dataloader(self) -> DataLoader:
         """Construct val dataloader"""
-        return DataLoader(self.val_dataset, shuffle=False, **self._common_dataloader_kwargs)
+        return DataLoader(
+            self.val_dataset,
+            shuffle=False,
+            drop_last=True,
```
dfulu: I think this should be false on the val dataloader

felix-e-h-p: Of course, just fixed!

dfulu (approved):
Thanks @felix-e-h-p

PR #532, opened 2026-06-24
Explicit error if devices: auto & >1 GPU available and if devices >1

bsharma58:
# Pull Request

## Description

Summary of changes:

- Check available CUDA GPU count before rejecting trainer configs.
- Reject `devices: "auto"` only when multiple GPUs are available.
- Reject explicit multi-GPU configs such as `devices: 2` and `devices: [0, 1]` when multiple GPUs are available.
- Allow safe configs such as `devices: 1`, `devices: [0]`, and `accelerator: cpu`.
- No extra dependencies required

Change motivated by non explicit error previously being raised and discussion during ML chat

## Fixes 

Testing:

- `uv run python -m pytest tests/models/test_validation.py`
- test validate gpu config rejects multi gpu configs when mutliple gpus available
- test validate gpu config allows configs when only one gpu available
- test validate gpu config allows single device configs

- [x] Yes

## Checklist:

### DIFF
--- a/pvnet/utils.py
+++ b/pvnet/utils.py
@@ -5,8 +5,9 @@ from typing import TYPE_CHECKING
 
 import rich.syntax
 import rich.tree
+import torch
 from lightning.pytorch.utilities import rank_zero_only
-from omegaconf import DictConfig, OmegaConf
+from omegaconf import DictConfig, ListConfig, OmegaConf
 
 if TYPE_CHECKING:
     from pvnet.models.base_model import BaseModel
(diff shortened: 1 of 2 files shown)


pvnet/utils.py
```diff
@@ -167,13 +168,48 @@ def validate_batch_against_config(
...
+    if accelerator in {"cpu", "mps", "tpu"}:
+        return
+
+    if devices == "auto":
+        msg = (
+            f"`devices: 'auto'` may use all {num_available_gpus} available GPUs. "
+            "PVNet does not support multi-GPU training. "
+            "Use a single explicit device, for example `devices: [0]`."
+        )
+    elif (
```
felix-e-h-p: Think this could be simplified to something like:

elif isinstance(devices, int) and devices > 1:

felix-e-h-p (changes requested):
Looking good, thanks for this! One small change if all OK.

tests/models/test_validation.py
```diff
@@ -23,21 +23,64 @@ def test_validate_batch_against_config_raises_error(late_fusion_model):
...
+    with pytest.raises(ValueError, match="does not support multi-GPU training"):
+        validate_gpu_config(trainer_cfg(trainer))
+
+@pytest.mark.parametrize(
+    "trainer",
+    [
+        {"devices": "auto"},
+        {"devices": 2},
+        {"devices": [0, 1]},
         {"accelerator": "cpu"},
```
felix-e-h-p: This also covered in test_validate_gpu_config_allows_single_device_configs right? Maybe redundant, but not 100% sure.

felix-e-h-p (changes requested):
Maybe one small change in tests also

pvnet/utils.py
bsharma58: Sweet, you're right the and not isinstance is futile here as true > 1 is false anyways!! updates now

tests/models/test_validation.py
bsharma58: Yep covered this in the last test :)), removed this now!

felix-e-h-p (approved):

PR #533, opened 2026-07-18
Updates

felix-e-h-p:
Enable TF32 tensor cores and make validation bf16-safe

- `run.py`: add `torch.set_float32_matmul_precision("high")` — lets fp32 matmuls / convs use tensor cores on Ampere+ GPUs. Free speedup, negligible precision cost.
- `lightning_module.py`: `y_hat.float().cpu().numpy()` in val storage - fixes a crash when training with `precision=bf16-mixed` (numpy has no bfloat16). No-op under fp32.

Tested on A6000 (DE model): GPU fully utilised, no change to training behaviour.

### DIFF
--- a/pvnet/training/lightning_module.py
+++ b/pvnet/training/lightning_module.py
@@ -153,6 +153,8 @@ class PVNetLightningModule(pl.LightningModule):
         """Internally store the validation predictions"""
 
         y = batch["generation"][:, -self.model.forecast_len :].cpu().numpy()
+        # .float() as numpy has no bfloat16 dtype
+        # Fails under precision="bf16-mixed"
         y_hat = y_hat.float().cpu().numpy()
         ids = batch["location_id"].cpu().numpy()
         init_times_utc = pd.to_datetime(
--- a/pvnet/training/lightning_module.py
+++ b/pvnet/training/lightning_module.py
@@ -153,7 +153,7 @@ class PVNetLightningModule(pl.LightningModule):
         """Internally store the validation predictions"""
 
         y = batch["generation"][:, -self.model.forecast_len :].cpu().numpy()
-        y_hat = y_hat.cpu().numpy()
+        y_hat = y_hat.float().cpu().numpy()
         ids = batch["location_id"].cpu().numpy()
         init_times_utc = pd.to_datetime(
             batch["time_utc"][:, self.model.history_len + 1].cpu().numpy().astype("datetime64[ns]")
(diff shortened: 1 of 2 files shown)


dfulu:
Looks good. What's the reasoning around the precision used? In pvnet-summation it was [locked to `"medium"`](https://github.com/openclimatefix/pvnet-summation/pull/46)

felix-e-h-p:
> Looks good. What's the reasoning around the precision used? In pvnet-summation it was [locked to `"medium"`](https://github.com/openclimatefix/pvnet-summation/pull/46)

Ah yeah, bit more precision than medium but gets same tensor core speedup on A6000s, maybe the safer default. Happy to change to medium to match the summation if you prefer.

dfulu:
So would that mean it would be better for the summation model to use `"high"` precision as well?

felix-e-h-p:
> So would that mean it would be better for the summation model to use `"high"` precision as well?

Not sure, but I suppose yeah - speed won't be a bottleneck anyway right. May be worth updating also.

dfulu:
Yeh it sounds like it would. It feels like we shouldn't have different standards for the two libraries

run.py
```diff
@@ -8,13 +8,15 @@
...
 import hydra
+import torch
 from omegaconf import DictConfig
 
 from pvnet.training import train
 from pvnet.utils import print_config, run_config_utilities, validate_gpu_config
 
 logging.basicConfig(stream=sys.stdout, level=logging.ERROR)
 
+torch.set_float32_matmul_precision("high")
```
dfulu, line 21: Can you add a comment about the reasoning for this

pvnet/training/lightning_module.py
```diff
@@ -153,7 +153,7 @@ def _store_val_predictions(self, batch: TensorBatch, y_hat: torch.Tensor) -> Non
         """Internally store the validation predictions"""
 
         y = batch["generation"][:, -self.model.forecast_len :].cpu().numpy()
-        y_hat = y_hat.cpu().numpy()
+        y_hat = y_hat.float().cpu().numpy()
```
dfulu, line 158: This could also do with a comment since its not obvious why this `.float()` call would be needed here

felix-e-h-p:
Couple comments added - also made the relevant update for summation: https://github.com/openclimatefix/pvnet-summation/pull/48

dfulu (approved):
Looks good, thanks

PR #536, opened 2026-07-28
Resume from ckpt path

### LINKED ISSUE #541: Resume training from checkpoint
If training crashes or fails, we might want to load from checkpoint and continue training. This would be a full state resume, including starting from the same optimizer state.

lm2612:
# Pull Request

## Description
We might want to continue training from same checkpoint e.g., if training run fails or crashes. This includes full state resume, including model weights and optimizer state. 

Fixes #541

I've added an option for `ckpt_path` in config. This should be the full path to the checkpoint that includes model weights, optimizer state, etc. This will be passed to the Lightning trainer in `pvnet/training/train.py`. If it is not specified, it will be ignored:

```python
trainer.fit(model=model, datamodule=datamodule, 
                ckpt_path=config.ckpt_path if 'ckpt_path' in config else None)
```

## How Has This Been Tested?

New test `test_checkpoint_load` in `tests/training/test_train.py`.  This will use the output from `test_train_pvnet` after the first epoch, and restart two training runs from there. Then we check that the saved checkpoints are identical. 

- [x] Yes

## Checklist:


tests/training/test_train.py
```diff
@@ -125,6 +128,78 @@ def test_train_pvnet(
...
+    assert len(ckpt_epoch1_path) == 2, f"expected 2 checkpoints at end of epoch 2, got {len(ckpt_epoch1_path)}, {ckpt_epoch1_path}"
+
+    # Load both checkpoints and compare 
+    ckpt0 = torch.load(ckpt_epoch1_path[0], map_location="cpu", weights_only=False)
+    ckpt1 = torch.load(ckpt_epoch1_path[1], map_location="cpu", weights_only=False)
+
+    # Compare state_dict
+    for key, value in ckpt0['state_dict'].items():
+        assert key in ckpt1['state_dict'],  f"model parameter {key} present in {ckpt_epoch1_path[0]} not found in {ckpt_epoch1_path[1]}"
+        assert ckpt1['state_dict'][key] == pytest.approx(value, abs=1e-9), f"model weights different for {key} by {ckpt1['state_dict'][key]-value}"
```
felix-e-h-p: Think can be reduced to:

torch.testing.assert_close(ckpt1['state_dict'][key], value, atol=1e-9, rtol=0)

tests/training/test_train.py
```diff
@@ -125,6 +128,78 @@ def test_train_pvnet(
...
+
+    # Check there are now two files at end of epoch=1
+    ckpt_epoch1_path = list(Path(wandb_save_dir).parent.glob("*/epoch=1*.ckpt"))
+    assert len(ckpt_epoch1_path) == 2, f"expected 2 checkpoints at end of epoch 2, got {len(ckpt_epoch1_path)}, {ckpt_epoch1_path}"
+
+    # Load both checkpoints and compare 
+    ckpt0 = torch.load(ckpt_epoch1_path[0], map_location="cpu", weights_only=False)
+    ckpt1 = torch.load(ckpt_epoch1_path[1], map_location="cpu", weights_only=False)
+
+    # Compare state_dict
```
felix-e-h-p, line 202: Think this loop and the optimizer state one could be collapsed into like a helper and then just a couple assert lines - not important but just to reduce lines a little

felix-e-h-p (changes requested):
All good, cheers! Just couple changes perhaps

tests/training/test_train.py
lm2612: thanks, just added a helper function to clean it up a bit

tests/training/test_train.py
lm2612: thanks! added

felix-e-h-p (approved):

Issue #542, opened 2026-08-20
save_last does not actually save the last checkpoint


Sukh-P:
## Describe the bug
Details of the bug can be found in this conversation [here](https://openclimatefix.slack.com/archives/C09AKPW1M7U/p1779200899491079) but essentially save_last actually saves a copy of the last checkpoint that was saved rather than the checkpoint for the last epoch. 

We should change this so we actually do save the last checkpoint, by e.g. either by having a customer saver or just saving out the checkpoint for the current epoch separately to the best one and keep overwriting so we keep the last one (might be done by a custom saver)

CodeVishal-17:
Hi! I'd like to work on this issue. I can reproduce the current `save_last` behavior, trace the checkpoint-saving flow, and implement the fix so that `last` always points to the checkpoint from the latest epoch. I'll also add/update regression tests to verify the behavior. Please assign this to me.

Sukh-P:
@CodeVishal-17 sure have at it but you may find this a bit hard to replicate since you will need to train `pvnet` with some open source dataset it supports to be able to actually see this issue/test the fix

CodeVishal-17:
@Sukh-P  Thanks for the clarification! That makes sense. I’ll skip this issue for now since reproducing and testing it requires the trained `pvnet` setup. Appreciate the explanation, and I’ll look for another issue that’s better suited for external contribution.

PR #544, opened 2026-09-10
save_last to save last checkpoint

felix-e-h-p:
Related to https://github.com/openclimatefix/pvnet/issues/542

- `save_last: True` was meant to save the model from the most recent epoch to `last.ckpt`.
- L 2.6.0 `last.ckpt` is written when a top-k checkpoint is saved at the same step.
- Used `save_top_k: 1`, so only happens when validation loss improves, and `last.ckpt` ends up being a copy of the best ckpt.
- Before 2.6, `save_last: True` did save real last epoch.

Changes:
- `configs.example/callbacks/default.yaml`: turned off `save_last`. added second `ModelCheckpoint` with `last_checkpoint`. Hence saves after every validation and overwrites `last.ckpt` each time.
- `pvnet/training/train.py`: Both callbacks now to be saved to the same run folder. The config files are still saved once, as before.
- `tests/training/test_train.py`: Updated fixture accordingly

### DIFF
--- a/configs.example/callbacks/default.yaml
+++ b/configs.example/callbacks/default.yaml
@@ -20,7 +20,7 @@ model_checkpoint:
   monitor: "${resolve_monitor_loss:${model.model.output_quantiles}}"
   mode: "min" # can be "max" or "min"
   save_top_k: 1 # save k best models (determined by above metric)
-  save_last: True # save a last.ckpt copy whenever a checkpoint file gets saved
+  save_last: False # save a last.ckpt copy whenever a checkpoint file gets saved
   every_n_epochs: 1
   verbose: False
   filename: "epoch={epoch}-step={step}"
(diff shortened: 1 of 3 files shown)


felix-e-h-p:
Should be marked as relevant issue in summation model also right?

configs.example/callbacks/default.yaml
```diff
@@ -20,11 +20,19 @@ model_checkpoint:
   monitor: "${resolve_monitor_loss:${model.model.output_quantiles}}"
   mode: "min" # can be "max" or "min"
   save_top_k: 1 # save k best models (determined by above metric)
-  save_last: True # save a last.ckpt copy whenever a checkpoint file gets saved
+  save_last: False # last.ckpt is saved by the last_checkpoint callback below
```
Sukh-P: Could we just not remove this field now? I think when removed it defaults to `None` so should have no effect?

Sukh-P:
@felix-e-h-p have you tested this already on a dummy model run? If so could you please link that wandb run and detail any checking you did to check this works as expected? Obviously no test coverage on this so would be good to know it's already been tried and works as expected

felix-e-h-p:
> @felix-e-h-p have you tested this already on a dummy model run? If so could you please link that wandb run and detail any checking you did to check this works as expected? Obviously no test coverage on this so would be good to know it's already been tried and works as expected

Yeah will do so later today

In here - https://wandb.ai/openclimatefix/pvnet_uk_test/runs/

Saving locally then gonna read step count and early stopping counter - have limited at low train and val batches so shouldn't be too long to check it

felix-e-h-p:
All good. 50 epochs with 100 train sample limit, best is 49, I loaded best and last just for a direct check: last step 5000 | best step 4900

Sukh-P (approved):

PR #546, opened 2026-09-14
Graceful fail for summation model in backtest

felix-e-h-p:
Relating to not using a summation model and no national i.e. 0 location in target - so specific regional backtests for NL. Graceful fail mode. Please correct me if wrong @bsharma58


scripts/backtest.py
```diff
@@ -156,6 +157,11 @@ def construct_model_data_config(
 class BacktestStreamedDataset(StreamedDataset):
     """A torch dataset object used only for backtesting"""
 
+    def __init__(self, *args, use_summation_model=False, **kwargs):
```
Sukh-P: I feel like maybe we don't need a default here for `use_summation_model`, we use `BacktestStreamedDataset` once and we can just have that field as required

Sukh-P:
I think this looks good @felix-e-h-p, wondering if it would be easy enough to run it without a summation model defined to check it's working properly

Sukh-P (approved):
Looks good but as mentioned would be great if this could be ran (with/without summation) just to double check it's all good since we lack tests for scripts

felix-e-h-p:
Ran on NL model (pvnet 6815348c), with summation (f2b9b22e) and without. Both runs fine.

